Issue #5, opened 2026-03-12
AI/LLM Integration Architecture Review


NikolayS:
### Scope
Review SPEC.md's AI integration (FR-10), session management (FR-15), autonomy model (FR-11), and prepare architecture for Phase 2 — with early decisions needed in Phase 0.

### Deliverables
- [ ] Context assembly pipeline design: what schema metadata to collect, compression strategy, refresh triggers
- [ ] Provider abstraction: trait/interface design that supports OpenAI, Anthropic, Ollama without vendor lock-in
- [ ] Token budget strategy: per-request limits, session limits, cost tracking, model selection
- [ ] Prompt template architecture for /ask, /explain, /fix, /optimize, /describe
- [ ] Session management: context compaction algorithm, conversation persistence, undo semantics
- [ ] Streaming response architecture: how LLM streaming integrates with TUI rendering
- [ ] Schema-aware context: what to include (tables, columns, indexes, constraints, stats), what to omit, dynamic context selection
- [ ] Plan mode / YOLO mode / observe mode: state machine design
- [ ] Early Phase 0 decisions: data model for schema cache, session storage (SQLite schema)
- [ ] Prompt injection risks in SQL generation context

### Context
Phase 2 is where Samo differentiates, but architecture decisions bleed backward. Schema context collection and session management must be designed before Phase 2 begins.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #35, opened 2026-03-13
feat(vars): variables, \pset, and output formats (CSV/JSON/HTML)

### LINKED ISSUE #32: feat(vars): \set/\unset variables and \pset output formatting
## Summary

Implement psql variable system (`\set`/`\unset`) and output format control (`\pset`) with all major output formats.

### Variables (`\set` / `\unset`)
- `\set [name [value]]` — set variable, or show all if no args
- `\unset name` — remove variable
- Built-in variables: AUTOCOMMIT, ON_ERROR_STOP, ON_ERROR_ROLLBACK, QUIET, ECHO, ECHO_HIDDEN, PROMPT1, PROMPT2, PROMPT3
- Variable interpolation in SQL: `:var`, `:'var'` (quoted), `:"var"` (double-quoted), `:{?var}` (boolean existence test)

### Output formatting (`\pset`)
- `\pset [option [value]]` — set output format option
- `\pset format` values: `aligned` (default), `unaligned`, `csv`, `json`, `html`, `wrapped`, `expanded`
- `\pset border` 0/1/2
- `\pset null` string
- `\pset fieldsep` / `\pset recordsep`
- `\pset tuples_only` on/off
- `\pset footer` on/off
- `\pset title` string
- `\pset expanded` on/off/auto

### Shortcut commands
- `\a` — toggle aligned/unaligned
- `\t [on|off]` — toggle tuples-only
- `\f [sep]` — set/show field separator
- `\H` — toggle HTML output
- `\C [title]` — set table title

### Output format implementations
Use/extend `src/output.rs` module:
- **CSV**: RFC 4180 compliant, comma-separated, quoted when needed
- **JSON**: array of objects with column names as keys
- **HTML**: `<table>` with `<thead>`/`<tbody>`
- **Unaligned**: field separator between columns, no padding
- **Wrapped**: aligned but wraps long values at terminal width

## Depends on
S-0.4 (metacmd parser) — merged

## Acceptance criteria
- [ ] `\set FOO bar` then `select :'FOO'` returns `'bar'`
- [ ] `\pset format csv` then `select 1 as a, 2 as b` outputs `a,b\n1,2`
- [ ] `\pset format json` outputs valid JSON array
- [ ] `\pset null [NULL]` displays NULL values as `[NULL]`
- [ ] `\a` toggles between aligned and unaligned
- [ ] Integration tests for CSV and JSON output

NikolayS:
## Summary

- New `src/vars.rs`: `Variables` struct with psql-compatible defaults and `interpolate()` supporting `:name`, `:'name'`, `:"name"`, `:{?name}`, with `::` cast passthrough
- `OutputFormat` enum + `PsetConfig` struct in `output.rs`; `ExpandedMode` moved there from `repl.rs`
- New formatters: RFC 4180 CSV, JSON array-of-objects, HTML `<table>`, unaligned (field-sep separated)
- New `MetaCmd` variants: `Set`, `Unset`, `Pset`, `ToggleAlign`, `TuplesOnly`, `FieldSep`, `ToggleHtml`, `SetTitle` with parser support
- `ReplSettings` extended with `pset: PsetConfig` and `vars: Variables`; `execute_query` interpolates variables before sending SQL
- CLI flags `--csv`, `--json`, `--no-align`, `--tuples-only`, `-F`, `-R`, `-z`, `-0`, `-P`, `-v` wired to initial `PsetConfig` / `Variables`

## Test plan

- [ ] `cargo fmt && cargo clippy -- -D warnings` — clean (verified)
- [ ] `cargo test` — 274 tests, 0 failures (verified)
- [ ] Variable set/get/unset unit tests in `vars::tests`
- [ ] Interpolation: `:var`, `:'var'`, `:"var"`, `:{?var}`, `::cast` passthrough
- [ ] CSV / JSON / HTML / unaligned format unit tests in `output::tests`
- [ ] `\set`, `\unset`, `\pset`, `\a`, `\t`, `\f`, `\H`, `\C` parser tests in `metacmd::tests`
- [ ] `samo -c "select 1 as a, 2 as b" --csv` → CSV output
- [ ] `samo -c "select 1 as a, 2 as b" --json` → JSON output
- [ ] `samo -c "\pset null [NULL]" -c "select null::text as val"` → shows `[NULL]`

Closes #32

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -141,9 +141,9 @@ struct Cli {
     #[arg(short = 't', long = "tuples-only")]
     tuples_only: bool,
 
-    /// Set printing option (like `\pset`).
+    /// Set printing option (like `\pset`). Can be specified multiple times.
     #[arg(short = 'P', long, value_name = "VAR[=ARG]")]
-    pset: Option<String>,
+    pset: Vec<String>,
 
     /// Send query results to file (or pipe).
     #[arg(short = 'o', long)]
@@ -290,12 +290,16 @@ fn apply_cli_pset(pset: &mut output::PsetConfig, arg: &str) {
     match option {
         "format" => {
             pset.format = match value.unwrap_or("") {
+                "aligned" => output::OutputFormat::Aligned,
                 "unaligned" => output::OutputFormat::Unaligned,
                 "csv" => output::OutputFormat::Csv,
                 "json" => output::OutputFormat::Json,
                 "html" => output::OutputFormat::Html,
                 "wrapped" => output::OutputFormat::Wrapped,
-                _ => output::OutputFormat::Aligned,
+                other => {
+                    eprintln!("samo: invalid value for -P format: \"{other}\"");
+                    std::process::exit(2);
+                }
             };
         }
         "border" => {
@@ -373,8 +377,8 @@ async fn main() {
                 sep.clone_into(&mut pset.record_sep);
             }
 
-            // Handle `\pset` from -P flag.
-            if let Some(ref pset_arg) = cli.pset {
+            // Handle `\pset` from -P flags (may be specified multiple times).
+            for pset_arg in &cli.pset {
                 apply_cli_pset(&mut pset, pset_arg);
             }
 
--- a/src/output.rs
+++ b/src/output.rs
@@ -132,9 +132,12 @@ impl Default for PsetConfig {
 
 /// Format a single [`RowSet`] using the active [`PsetConfig`].
 pub fn format_rowset_pset(out: &mut String, rs: &RowSet, cfg: &PsetConfig) {
-    // Title line (if set).
-    if let Some(ref title) = cfg.title {
-        let _ = writeln!(out, "{title}");
+    // Title line: printed as plain text for non-HTML formats.
+    // HTML format emits the title itself as <caption> inside the table element.
+    if cfg.format != OutputFormat::Html {
+        if let Some(ref title) = cfg.title {
+            let _ = writeln!(out, "{title}");
+        }
     }
 
     match &cfg.format {
@@ -671,6 +674,10 @@ fn csv_field(val: &str) -> String {
 /// Each row becomes `{"col1": "val1", "col2": "val2"}`.
 /// NULL values are rendered as JSON `null`.
 /// String values are JSON-escaped.
+///
+/// `tuples_only` is intentionally ignored: JSON output always includes column
+/// keys because removing them would produce invalid/ambiguous data (an array of
+/// bare values with no key context).  This matches psql behaviour.
 pub fn format_json(out: &mut String, rs: &RowSet, _cfg: &PsetConfig) {
     let cols = &rs.columns;
     let rows = &rs.rows;
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -668,7 +668,7 @@ fn apply_unset(settings: &mut ReplSettings, name: &str) {
 /// Apply a `\pset` command.
 #[allow(clippy::too_many_lines)]
 fn apply_pset(settings: &mut ReplSettings, option: &str, value: Option<&str>) {
-    use crate::output::{OutputFormat, PsetConfig};
+    use crate::output::OutputFormat;
 
     if option.is_empty() {
         // Display all pset options.
@@ -678,6 +678,11 @@ fn apply_pset(settings: &mut ReplSettings, option: &str, value: Option<&str>) {
 
     match option {
         "format" => {
+            if value.is_none_or(str::is_empty) {
+                // \pset format (no value) — show current setting.
+                println!("Output format is {}.", format_name(&settings.pset.format));
+                return;
+            }
             let fmt = match value.unwrap_or("") {
                 "aligned" => OutputFormat::Aligned,
                 "unaligned" => OutputFormat::Unaligned,
@@ -764,7 +769,6 @@ fn apply_pset(settings: &mut ReplSettings, option: &str, value: Option<&str>) {
 
     // Keep ReplSettings.expanded in sync.
     settings.expanded = settings.pset.expanded;
-    let _ = PsetConfig::default; // suppress unused-import lint
 }
 
 /// Parse a boolean value for pset options: `on`/`true`/`1` → true, else toggle.


NikolayS (commented):
Good overall architecture — splitting `vars.rs` out, moving `ExpandedMode` to `output.rs`, and the `PsetConfig`/`format_rowset_pset` approach are all clean. The test coverage is solid. That said, I found several bugs and a few noteworthy design issues.

---

## Bugs

### 1. HTML title rendered outside `<table>` — invalid HTML structure

`format_rowset_pset` (output.rs) prints the title as a plain text line **before** calling `format_html`. `format_html` then **also** checks `cfg.title` and emits a `<caption>` inside the table. This produces:

```
My Title          ← from format_rowset_pset (plain text)
<caption>My Title</caption>   ← from format_html
<table>
…
```

The `<caption>` must be the first child of `<table>`, not preceded by raw text. And the title is printed twice. Fix: in `format_rowset_pset`, skip the title writeln when format is HTML (let `format_html` handle it entirely), or remove the `cfg.title` block from `format_html` and let the caller always handle it.

```rust
// output.rs – format_rowset_pset
if let Some(ref title) = cfg.title {
    let _ = writeln!(out, "{title}");  // ← emitted for ALL formats, including HTML
}
```

---

### 2. `apply_pset` runs the `expanded` sync unconditionally — corrupts state on early-return

`apply_pset` in repl.rs ends with:

```rust
// Keep ReplSettings.expanded in sync.
settings.expanded = settings.pset.expanded;
```

This line runs even when `apply_pset` returned early (unknown format → `eprintln!` + `return`, unknown option → `other` arm, or the `border` error path). Those early returns all fall through to this sync. For the `other` arm and border error, the early `return` is **before** the sync, so they're actually fine — but any future handler that returns early without reaching this line will silently desync. More importantly:

The `format` arm does `return` after printing (it doesn't fall through), so `settings.expanded` is **never synced after `\pset format csv`**. This isn't wrong today (format doesn't touch expanded) but the architecture is fragile. Consider syncing at the start of the function or making it explicit per-arm.

---

### 3. `\set name value` — multi-word values are trimmed but leading whitespace in the remainder is consumed incorrectly

`parse_set` uses `splitn(2, char::is_whitespace)` which splits on the **first** whitespace character. The second half is then `.trim()`-ed. This means `\set X  hello` (two spaces) correctly gives `"hello"`, but `\set X  hello world` gives `"hello world"` (single trim). That's correct. However:

`splitn(2, char::is_whitespace)` matches on **any single** whitespace char. When the separator is a tab (`\set X\thello`), the split works fine. But the **name** extraction via `splitn` can produce a name that itself contains a tab if the user writes something unusual. This is a minor edge case; the real psql also has quirks here. No action required unless psql-compatibility is strict.

---

### 4. `::` passthrough is correct, but `:::name` will misfire

Given `select 1:::myvar` (three colons, rare but valid in some contexts), the current logic:
- Sees `:` at position i
- Sees `:` at i+1 → emits `::`, advances i by 2
- Now at position i+2, sees `:` followed by `m` → interpolates `:myvar`

Result: `select 1::42` — the three-colon sequence is treated as a cast + interpolation. This matches psql's actual behavior, so it's correct. Just noting it's tested implicitly by `interp_cast_in_complex_query`.

---

### 5. `apply_cli_pset` silently ignores `format=aligned` (maps unknown → Aligned)

In `apply_cli_pset` (main.rs):

```rust
"format" => {
    pset.format = match value.unwrap_or("") {
        "unaligned" => output::OutputFormat::Unaligned,
        ...
        _ => output::OutputFormat::Aligned,   // ← "aligned" hits this arm, so it works
    };
}
```

The `_ =>` arm catches both `"aligned"` (correct) and any **typo** like `"alignedd"` or `"csvv"` without any error. In contrast, `apply_pset` (repl.rs) correctly rejects unknown formats with `eprintln!`. The CLI path should match:

```rust
other => {
    eprintln!("samo: unknown format \"{other}\"");
    return; // or continue with default
}
```

---

### 6. `\set name` (display-only) is ambiguous with `\set name ""` (set to empty string)

Both `\set FOO` and `\set FOO ` after trimming produce `value = ""`. The code distinguishes them by checking `value.is_empty()`:

```rust
if value.is_empty() {
    // Display one variable.
    ...
    return;
}
settings.vars.set(name, value);
```

This means **you cannot set a variable to an empty string** via `\set FOO ` (trailing space). In psql, `\set FOO` (no value) displays the variable, but `\set FOO ''` would be needed to set it empty. Since `\set` doesn't support shell-style quoting, this limitation is acceptable **but should be documented** — currently the code has no comment explaining the empty-string restriction.

---

### 7. `format_csv` does not handle `null_display` containing special characters

When `null_display` contains a comma or quote, `csv_field` is called correctly (it's used via `v.as_deref().unwrap_or(&cfg.null_display)`). Actually this path **is** correct — `csv_field` is called on the unwrapped value. Wait — the null case *does* go through `csv_field`:

```rust
.map(|v| csv_field(v.as_deref().unwrap_or(&cfg.null_display)))
```

This is fine. No bug here. Struck from the list.

---

### 8. `format_unaligned` footer — uses `writeln!` but rest of function uses `push_str`

Minor inconsistency: the header and data use `out.push_str(&cfg.record_sep)` (no guaranteed newline), but the footer uses `writeln!` (always `\n`). If `record_sep` is set to `\0` (NUL, via `-0`), the data rows end with NUL but the footer ends with `\n`. psql emits the row-count footer with `\n` regardless of record separator, so this is actually **correct psql behavior** — just worth a comment.

---

### 9. `-P` only accepts a single `pset` option (clap `Option<String>`)

psql allows `-P` to be specified **multiple times** (e.g., `psql -P format=csv -P null=[NULL]`). The current definition:

```rust
#[arg(short = 'P', long, value_name = "VAR[=ARG]")]
pset: Option<String>,
```

Only the last `-P` would be accepted (actually clap would error on duplicate). Should be `Vec<String>` with the apply loop iterating all values.

---

### 10. `\pset` with `format` and no value — falls through to unknown format error

```rust
"format" => {
    let fmt = match value.unwrap_or("") {
        "aligned" => ...
        ...
        other => {
            eprintln!("\\pset: unknown format \"{other}\"");
            return;
        }
    };
```

`\pset format` (no value) produces `value = None`, so `unwrap_or("")` gives `""`, which hits `other => eprintln!(... "")`. psql with `\pset format` (bare) displays the current format. This is a minor UX divergence, not a correctness bug, but worth noting.

---

## Minor / Code Quality

- **`let _ = PsetConfig::default;`** (repl.rs line 767): This is a no-op used to "suppress an unused-import lint" — but `PsetConfig` is already imported at the top of `apply_pset` via `use crate::output::{OutputFormat, PsetConfig}`. The lint suppression is unnecessary and confusing. Remove it.

- **`_cfg` parameter in `format_json`**: The parameter is `_cfg: &PsetConfig` (prefixed with `_` to suppress the warning). If `tuples_only` should suppress header for JSON (it doesn't apply, since JSON is always an array), this is fine. But if a future caller passes `tuples_only: true` expecting rows-only JSON, there's no effect. A brief comment explaining the intentional `_cfg` would help.

- **No test for `-P format=csv` CLI path in `apply_cli_pset`**: The `apply_cli_pset` function has no unit tests. Given it has the silent-fallthrough-to-Aligned bug (#5 above), a test would catch this.

- **`\set FOO bar baz` parses value as `"bar baz"`** (trimmed): The test `parse_set_value_with_spaces` covers `\set X hello world` → `"hello world"`. But the trimming is done with `str::trim`, which only trims the **leading and trailing** whitespace of the second half. `\set X  bar  baz  ` would give `"bar  baz"` (internal spaces preserved, trailing stripped). This matches psql.

---

## Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **Bug** | `output.rs:format_html` + `format_rowset_pset` | Title printed twice, `<caption>` outside `<table>` |
| 2 | Minor | `repl.rs:apply_pset` | Unconditional expanded-sync after early returns (fragile) |
| 5 | **Bug** | `main.rs:apply_cli_pset` | Unknown format silently mapped to Aligned instead of error |
| 9 | **Bug** | `main.rs` CLI | `-P` is `Option<String>`, should be `Vec<String>` for multiple uses |
| 10 | Minor UX | `repl.rs:apply_pset` | `\pset format` (bare) prints error instead of showing current format |

PR #356, opened 2026-03-14
fix(config): use Option instead of sentinel values for border/pager_min_lines (M3)

### LINKED ISSUE #339: [REV] Code Review — v0.2.0 (da4c1d8)
# Code Review — Samo v0.2.0

**Scope:** Full codebase as of `da4c1d8` (46K lines, 42 source files)
**Build:** ✅ Clean clippy, 1354 tests passing, zero warnings

---

## Summary

The codebase has matured significantly from the initial 31K-line build. New additions — SSH tunnel support, daemon mode with anomaly detection, markdown rendering, project config, log rotation, compat report — are all well-structured with good test coverage. The SQLite dependency was already removed (now JSON file-backed sessions). Code quality is high: no `unsafe` outside of one legitimate `localtime_r` call, no panics outside tests, all `unwrap()` calls in test code.

**Verdict:** Production-quality for a v0.2 milestone. The issues below are refinements, not blockers.

---

### C1. AI config: no provider auto-detection from `api_key_env`

**File:** `src/repl.rs:6901-6913`

The "AI not configured" check looks at `config.ai.provider`, which defaults to `None`. If a user writes:

```toml
[ai]
api_key_env = "OPENAI_API_KEY"
```

...without explicitly setting `provider = "openai"`, they get the "not configured" error even though the intent is obvious. The env var name contains the provider name. This is the bug Nik reported.

**Fix:** Infer provider from `api_key_env` when `provider` is `None`:
- `OPENAI_API_KEY` / `OPENAI_*` → `"openai"`
- `ANTHROPIC_API_KEY` / `ANTHROPIC_*` → `"anthropic"`
- `OLLAMA_*` → `"ollama"`

Do this in `load_config()` or as a post-load step in `Config`, not scattered across 9+ call sites.

### C2. `chrono_now()` returns epoch seconds, not ISO 8601

**File:** `src/daemon.rs:335-342`

The doc comment says "Get current time as ISO 8601 string" but the function returns raw epoch seconds (`"1710373401"`). The `HealthStatus` struct also documents `last_check` as ISO 8601. The health endpoint test (`health_status_json_connected`) manually constructs an ISO string, so it passes — but the real runtime would emit epoch seconds.

The health JSON endpoint becomes useless for humans/monitoring when `last_check` is an opaque number.

**Fix:** Format as `YYYY-MM-DDTHH:MM:SSZ`. The logging module already has `current_time_hms()` — extend it or write a proper UTC formatter. No need for the `chrono` crate; basic arithmetic on epoch seconds works.

---

### H1. `repl.rs` is 10,653 lines — decomposition overdue

343 functions in one file. The AI command handlers alone (`handle_ai_ask`, `handle_ai_plan`, `handle_ai_fix`, `handle_ai_explain`, `handle_ai_optimize`, `handle_ai_describe`, `handle_ai_rca`) are ~1500 lines that share no state with the REPL loop itself. The `\watch` implementation, editor integration, and prompt rendering are also independent.

**Recommendation:** Extract into modules:
- `src/repl/ai_commands.rs` (~1500 lines)
- `src/repl/watch.rs`
- `src/repl/editor.rs`
- `src/repl/prompt.rs`
- `src/repl/mod.rs` (the REPL loop + dispatch)

This doesn't change any public API — it's purely internal refactoring.

### H2. `resolve_api_key()` duplicated at 9 call sites

Every AI command independently calls `resolve_api_key()` and `create_provider()`. This is ~15 lines of identical boilerplate repeated 9 times. If the provider setup logic changes (e.g., adding model validation, rate limiting, connection pooling), you need to update 9 places.

**Fix:** Extract a `fn get_ai_provider(settings) -> Result<Box<dyn AiProvider>, String>` helper that handles the full chain: check provider → resolve key → create provider → check token budget. Call it from one place.

### H3. SSH tunnel accepts all host keys

**File:** `src/ssh_tunnel.rs:134`

`check_server_key` returns `Ok(true)` unconditionally — equivalent to `StrictHostKeyChecking=no`. This is a MITM vector. The TODO acknowledges it.

For v0.2 this is acceptable with a warning, but before any production recommendation:
1. Print a visible warning when connecting ("WARNING: host key verification disabled")
2. Track as a blocking issue for v0.3

### H4. Health endpoint has no authentication

**File:** `src/daemon.rs:156-183`

The HTTP health server binds to `127.0.0.1:{port}` and responds to any request. While localhost-only, on shared hosts any user can hit it. The endpoint leaks:
- Database name
- Connection status
- Active anomaly count

Not critical for localhost, but worth noting. Consider adding a simple bearer token check for non-localhost binds, or document that it should only be exposed via a reverse proxy.

---

### M1. Slack notification body is not properly escaped

**File:** `src/daemon.rs:101`

```rust
let payload = format!(r#"{{"text":"{}"}}"#, message.replace('"', r#"\""#));
```

This only escapes double quotes. Slack JSON payloads also need `\n`, `\t`, `\r`, and `\\` escaped. If an anomaly description contains a newline or backslash, the JSON is malformed and the notification silently fails.

**Fix:** Use `serde_json::json!({ "text": message })` — `serde_json` is already a dependency.

### M2. `mask_credentials` doesn't handle URL-encoded passwords

**File:** `src/logging.rs:410`

```rust
if let Some(idx) = lower.find("password=") { ... }
```

Postgres connection URIs use `postgresql://user:p%40ss@host/db`. The password is between `:` and `@`, not after `password=`. The function would miss it entirely in URI format.

### M3. Config merge: `display.border` uses `1` as sentinel for "not set"

**File:** `src/config.rs:803`

```rust
border: if overlay.display.border == 1 { base.display.border } else { overlay.display.border },
```

If a user explicitly sets `border = 1` in their config, it's treated as "not set" and the base value wins. Same issue exists for `pager_min_lines` with `0` as sentinel. Use `Option<u8>` for these fields instead.

### M4. Test for `check_existing_pid` uses `kill -0` via `Command::new("kill")`

**File:** `src/daemon.rs:81`

Shelling out to `kill -0` is fragile. Use `libc::kill(pid as i32, 0)` directly — it's a single syscall, already available via the `libc` crate in `Cargo.toml`.

---

## 📊 Stats

| Metric | Value |
|--------|-------|
| Source files | 42 |
| Total lines | 45,859 |
| Largest file | `repl.rs` (10,653) |
| Dependencies | 26 (runtime) |
| Tests | 1,354 |
| Clippy warnings | 0 |
| `unsafe` blocks | 1 (legitimate `localtime_r`) |
| `unwrap()` in non-test code | 0 (all in `writeln!` to `String`, which cannot fail) |

---

*Reviewed by Leo (automated REV) at da4c1d8*

NikolayS:
## Summary

Addresses M3 from code review issue #339.

- Replace `DisplayConfig.border: u8` with `Option<u8>` — the old sentinel `1` made it impossible for a user to explicitly set `border = 1` in an overlay config (it was treated as "not set")
- Replace `DisplayConfig.pager_min_lines: usize` with `Option<usize>` — the old sentinel `0` meant setting `pager_min_lines = 0` in a project config was silently ignored
- Merge logic now uses `Option::or()` — clean and correct, no sentinel comparisons
- In `main.rs`: apply config-file `border` _before_ CLI `-P` args so CLI always wins without needing sentinel comparisons; remove the old heuristic `if pset.border == 1 && cfg.display.border != 1` block
- `pager_min_lines` resolved via `.unwrap_or(0)` at the point of use (default behaviour preserved)
- Two new regression tests: `merge_display_border_base_preserved_when_overlay_none` and `merge_display_border_explicit_one_overlay_wins` (the latter explicitly documents the bug that was impossible to test before)

## Test plan

- [x] `cargo build` — clean
- [x] `cargo fmt -- --check` — clean
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo test` — 1366 passed, 0 failed

Closes #339 (M3)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -95,10 +95,16 @@ pub struct DisplayConfig {
     pub timing: bool,
     /// Expanded display mode (like `\x`). Default: `false`.
     pub expanded: bool,
-    /// Minimum output lines before the pager activates. Default: `0` (always).
-    pub pager_min_lines: usize,
-    /// Table border style (`0`, `1`, or `2`). Mirrors `\pset border`. Default: `1`.
-    pub border: u8,
+    /// Minimum output lines before the pager activates.
+    ///
+    /// When `None` (or absent from config), the default `0` applies (always
+    /// page).  When `Some(n)`, the pager only activates if output exceeds `n`
+    /// lines.
+    pub pager_min_lines: Option<usize>,
+    /// Table border style (`0`, `1`, or `2`). Mirrors `\pset border`.
+    ///
+    /// When `None` (or absent from config), the default `1` applies.
+    pub border: Option<u8>,
     /// Use Vi keybinding mode in the REPL. Default: `false` (Emacs mode).
     ///
     /// When `true`, rustyline uses `EditMode::Vi` instead of the default
@@ -130,8 +136,8 @@ impl Default for DisplayConfig {
             highlight: true,
             timing: false,
             expanded: false,
-            pager_min_lines: 0,
-            border: 1,
+            pager_min_lines: None,
+            border: None,
             vi_mode: false,
             // Default ON — overridden to OFF in non-interactive sessions.
             statusline_enabled: true,
@@ -850,16 +856,11 @@ fn merge_config(base: Config, overlay: Config) -> Config {
             highlight: overlay.display.highlight,
             timing: overlay.display.timing,
             expanded: overlay.display.expanded,
-            pager_min_lines: if overlay.display.pager_min_lines == 0 {
-                base.display.pager_min_lines
-            } else {
-                overlay.display.pager_min_lines
-            },
-            border: if overlay.display.border == 1 {
-                base.display.border
-            } else {
-                overlay.display.border
-            },
+            pager_min_lines: overlay
+                .display
+                .pager_min_lines
+                .or(base.display.pager_min_lines),
+            border: overlay.display.border.or(base.display.border),
             vi_mode: overlay.display.vi_mode || base.display.vi_mode,
             // Prefer explicit false from overlay over base default.
             statusline_enabled: overlay.display.statusline_enabled
@@ -981,8 +982,8 @@ mod tests {
         assert!(cfg.display.highlight); // default
         assert!(!cfg.display.timing);
         assert!(!cfg.display.expanded);
-        assert_eq!(cfg.display.pager_min_lines, 0);
-        assert_eq!(cfg.display.border, 1);
+        assert_eq!(cfg.display.pager_min_lines, None);
+        assert_eq!(cfg.display.border, None);
         assert!(!cfg.display.vi_mode); // default is Emacs
         assert!(cfg.safety.destructive_warning);
         assert!(cfg.connection.host.is_none());
@@ -1061,8 +1062,8 @@ pager_min_lines = 40
 border = 2
 ";
         let cfg: Config = toml::from_str(toml_str).expect("should parse");
-        assert_eq!(cfg.display.pager_min_lines, 40);
-        assert_eq!(cfg.display.border, 2);
+        assert_eq!(cfg.display.pager_min_lines, Some(40));
+        assert_eq!(cfg.display.border, Some(2));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


NikolayS:
Duplicate — superseded by #362.

PR #594, opened 2026-03-17
fix(ai): stop /fix loops on identity column errors in t2s mode

NikolayS:
## Summary

- **Schema context**: `build_schema_context` now queries `pg_attribute.attidentity` and annotates identity columns with `generated always as identity` or `generated by default as identity` in the DDL stubs shown to the LLM, so it knows those columns must be omitted from `INSERT` lists.
- **System prompt**: the `/fix` system prompt now has two explicit rules: never include identity columns in `INSERT` column lists, and never use `nextval()` as a workaround — just omit them.
- **Fix history**: `handle_ai_fix` previously built a fresh two-message context on every call, so the AI had no memory of what it had already tried. It now prepends the full conversation history before the AI call and records each (error, suggestion) pair into the conversation context, breaking the hallucination loop.

## Root cause (issue #591)

The AI generated `INSERT INTO orders (id, ...) SELECT generate_series(1,1000000) AS id, ...` where `id` is `GENERATED ALWAYS AS IDENTITY`. Repeated `/fix` calls each got a blank context and kept hallucinating sequences (`orders_id_seq`) that don't exist, spiralling into more errors. The correct fix — simply omit `id` from the column list — is now in the system prompt and schema annotations.

## Test plan

- [ ] `cargo test ai::context` — all 11 tests pass (3 new identity-column tests added)
- [ ] `cargo clippy` — zero warnings in new code
- [ ] `cargo build` — clean compile
- [ ] Manually verify: table with `GENERATED ALWAYS AS IDENTITY` pk → `/ask` insert query → AI omits the identity column
- [ ] Manually verify: after a failed INSERT on identity column, `/fix` produces correct omission, not sequence reference
- [ ] Manually verify: repeated `/fix` after a failing fix now shows different suggestions rather than looping

### DIFF
--- a/src/repl/ai_commands.rs
+++ b/src/repl/ai_commands.rs
@@ -1859,7 +1859,7 @@ mod tests {
 
     // -- ConversationContext loop-prevention mechanism -------------------------
 
-    /// Verify that messages pushed into ConversationContext are returned by
+    /// Verify that messages pushed into `ConversationContext` are returned by
     /// `to_messages()` in order and with the correct roles.  This is the
     /// mechanism that injects prior /fix attempts into subsequent AI calls,
     /// preventing the AI from suggesting the same wrong fix repeatedly.


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/rpg#594 — fix(ai): stop /fix loops on identity column errors in t2s mode
- **Author:** NikolayS
- **CI Status:** ✅ all checks passed

---

### BLOCKING ISSUES (1)

**MEDIUM** `src/repl/ai_commands.rs:handle_ai_fix` — Missing `auto_compact_if_needed()` call after pushing to conversation *(confidence 8)*
> `handle_ai_fix` now records each (error, suggestion) pair into `settings.conversation` to prevent repeated wrong fixes — but unlike `handle_ai_ask` (which calls `auto_compact_if_needed()` after every push), `handle_ai_fix` never compacts. In a repeated `/fix` loop (the exact scenario this PR targets), conversation history grows unboundedly: each entry includes the full error, full SQL, and full AI response. This can eventually exceed the model's context window, causing API errors and defeating the purpose of carrying history.
> **Fix:** Add after the push calls in `handle_ai_fix`:
> ```rust
> settings.conversation.auto_compact_if_needed(settings.config.ai.context_window);
> ```

---

### POTENTIAL ISSUES (2)

**LOW** Prior conversation history (including prior `/ask` exchanges) is injected into the `/fix` prompt *(confidence 6)* — `/fix` and `/ask` share `settings.conversation`, so unrelated general chat pollutes the error-correction context and vice versa. Consider tagging entries or using separate contexts.

**LOW** Conversation history introduces a self-injection surface *(confidence 7)* — crafted SQL or error messages in prior turns could contain adversarial instructions for the LLM. Low risk in single-user local REPL; flag if extended to automated/multi-user contexts.

---

### Summary

| Area | Blocking | Potential | Filtered |
|------|----------|-----------|---------|
| Bugs | 1 | 1 | 0 |
| Security | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |

**Result: BLOCKED — 1 blocking issue must be fixed before merge.**

---
*REV-assisted review (AI analysis)*

NikolayS:
## REV Code Review — PR #594

> `fix(ai): stop /fix loops on identity column errors in t2s mode`
>
> Five parallel agents: security (opus-level), bugs (opus-level), tests (sonnet), guidelines (sonnet), docs (sonnet).

---

### Summary

| Category | BLOCKING | NON-BLOCKING | POTENTIAL |
|----------|----------|--------------|-----------|
| Security | 0 | 1 (SEC-2) | 1 (SEC-1) |
| Bugs | 1 (BUG-2) | 1 (BUG-3) | 1 (BUG-4) |
| Tests | 1 (TEST-4) | 2 (TEST-2, TEST-5) | 1 (TEST-3) |
| Guidelines | 0 | 3 (GUIDE-1,2,7) | 0 |
| Docs | 0 | 1 (DOC-4) | 1 (DOC-5) |
| **Total** | **2** | **8** | **4** |

**Recommendation: DO NOT MERGE until BUG-2 and TEST-4 are resolved.**

---

### BLOCKING Findings

#### BUG-2 — Conversation records pre-edit suggestion, not what was executed (MEDIUM, confidence 8)

**File:** `src/repl/ai_commands.rs`

After the AI returns a fix, `push_user(user_content)` and `push_assistant(result.content)` are called unconditionally — before the user chooses Y/n/e. When the user selects **Edit** (`AskChoice::Edit`), they modify the SQL and execute a different version than what the AI suggested. On the next `/fix` call, the conversation history will show the AI's original (potentially wrong) suggestion as what was tried, not the user-edited version. This can mislead the AI into thinking its original suggestion was executed and failed, causing it to diverge further.

**Suggestion:** Move `push_user`/`push_assistant` to after the execution decision, or record the actually-executed SQL (including edits) as the history entry. For the `AskChoice::Edit` path, push the edited SQL variant. For `AskChoice::No`, consider whether to record the declined suggestion at all.

---

#### TEST-4 — No tests for the core behavioral change: conversation history in /fix (HIGH, confidence 8)

**File:** `src/repl/ai_commands.rs`

The most significant change in this PR — `/fix` now reads from and writes to `settings.conversation` — has zero unit or integration tests. `ConversationContext` methods are tested in `mod.rs`, but there is no test verifying that `handle_ai_fix`:

- (a) prepends existing conversation entries before the AI call
- (b) records the user turn and assistant turn after completion
- (c) triggers `auto_compact_if_needed`
- (d) handles the case where user declines/edits (no double-recording)

The PR test plan relies on manual verification for this behavior. Since this is the fix for the hallucination loop (the entire point of the PR), the core logic path needs automated coverage.

**Suggestion:** Extract the message-building logic from `handle_ai_fix` into a pure function that can be unit-tested. Alternatively, add an integration test using a mock `LlmProvider` that verifies the message slice seen by the AI on a second `/fix` call contains the prior turn.

---

### NON-BLOCKING Findings

#### SEC-2 — Sensitive data accumulation in conversation context sent to LLM API (LOW, confidence 8)

`handle_ai_fix` now shares `settings.conversation` with `/ask`. Query results and error messages that accumulate from `/ask` sessions will be re-sent to the external LLM provider on every `/fix` call. `auto_compact` truncates to 200-char previews but full-fidelity data persists until compaction. This is consistent with existing `/ask` behavior, but worth documenting in a user-facing note (e.g., in help text or docs) so users understand what is sent externally.

---

#### BUG-3 — Auto-compact after /fix may lose exact error context (INFO, confidence 8)

When `auto_compact_if_needed` triggers during a `/fix` loop, older error messages are summarized to 200-char previews. The AI loses access to exact prior error text, which may reduce fix accuracy on the next iteration. This is the design intent of compaction, but it creates a subtle tension: the fix that prompted the conversation-history feature (to break hallucination loops) can itself be degraded by compaction. Consider whether the `/fix`-specific entries should be marked `is_action=true` to survive compaction.

---

#### GUIDE-1 — PR title exceeds 50-character limit (LOW, confidence 10)

Title is 62 characters: `"fix(ai): stop /fix loops on identity column errors in t2s mode"`. CLAUDE.md: *"Subject < 50 chars"*.

Suggested: `fix(ai): stop /fix loops on identity columns` (45 chars)

---

#### GUIDE-2 — Pre-existing uppercase SQL keywords in WAIT_SNAPSHOT_SQL (LOW, confidence 9)

`WAIT_SNAPSHOT_SQL` and `PG_ASH_RECENT_WAITS_SQL` in `context.rs` use uppercase keywords (SELECT, FROM, WHERE, GROUP BY, ORDER BY). CLAUDE.md requires lowercase. Not introduced by this PR, but the PR touches `context.rs` — a good opportunity to fix. File: `src/ai/context.rs` lines 97–121.

---

#### DOC-4 — handle_ai_fix doc comment not updated for new statefulness (LOW, confidence 8)

`handle_ai_fix` now reads and writes `settings.conversation`. If a function-level doc comment exists for this function, it should document this side effect. The inline comments within the function body are good, but the function signature-level documentation should also note that conversation history is consumed and updated.

---

### POTENTIAL Findings

#### SEC-1 — Prompt injection via conversation history (MEDIUM, confidence 7)

DB error messages (which now accumulate in conversation history and are re-sent as User role messages to the LLM) could in theory contain injection payloads. Since these are user-owned databases the attack surface is self-injection only, and User-role placement reduces risk vs System-role. Acceptable trade-off given the fix scope, but worth tracking.

#### BUG-4 — /ask and /fix share conversation context without isolation (MEDIUM, confidence 7)

`handle_ai_fix` writes to `settings.conversation`, the same context used by `/ask`. A long `/ask` session before `/fix` injects unrelated prior queries into the fix context. Consider a dedicated `fix_conversation` or filtering entries by command type.

#### TEST-3 — Non-PK identity column not tested (LOW, confidence 7)

`identity_kind='a'` with `is_pk=false` produces `"col type generated always as identity,"` — a valid PG configuration not covered by the new tests.

#### DOC-5 — Compaction trade-off not noted near auto_compact call (INFO, confidence 7)

The inline comment at the `auto_compact_if_needed` call in `handle_ai_fix` doesn't note that compaction may reduce fix accuracy by summarizing error history. A brief TODO would help future maintainers.

---

### Positive Notes

- The root cause analysis (issue #591) is accurate and well-documented.
- `pg_catalog.pg_attribute.attidentity` values (`'a'`, `'d'`, `''`) are PostgreSQL-system-controlled — no injection risk from that column (SEC-3 clear).
- The three new identity column tests (`identity_always_column_annotated`, `identity_by_default_column_annotated`, `non_identity_column_has_no_identity_annotation`) are well-structured and cover all match arms.
- All 8 existing `context.rs` tests are correctly updated to the new 6-tuple signature with no regressions.
- The system prompt additions are clear and actionable for the LLM.
- SQL style in the new catalog query (`a.attidentity as identity_kind`) is correct: lowercase, uses `AS`, proper formatting.

---

🤖 REV review by Claude Sonnet 4.6 — 5 agents (security, bugs, tests, guidelines, docs)

PR #596, opened 2026-03-17
fix(repl): yolo auto-enables text2sql; fix \echo escapes and \set concat

NikolayS:
## Summary

- **Bug 1 (`\yolo`):** `\yolo` now also sets `input_mode = Text2Sql`, so natural language typed after `\yolo` is routed to the AI instead of the SQL buffer. All three `MetaResult::SetExecMode` dispatch sites updated.

- **Bug 2 (`\echo` escape sequences):** Added `unescape_echo()` helper that processes `\n`, `\t`, `\r`, `\b`, `\f`, `\\`, `\'`, `\ooo` (octal), and `\xhh` (hex) — matching psql. The `\echo` handler now calls `split_params` (strips surrounding single-quotes, handles `\'`/`''` inside quoted strings) then joins tokens with spaces, then calls `unescape_echo`. This makes `\echo '\033[1;35mMenu:\033[0m'` emit real ANSI colour codes, fixing postgres_dba coloured menus.

- **Bug 3 (`\set` token concatenation):** `split_params` is now `pub` and supports `\'` as a quote escape inside single-quoted strings (in addition to the existing `''` doubled-quote). `parse_set` now runs the value part through `split_params` and concatenates tokens without spaces, matching psql. This fixes `\set d '\'' :var '\''` → `d = 'value'` as used by postgres_dba.

## Test plan

- [ ] `cargo test` — 1391 tests, all pass
- [ ] `cargo clippy` — no warnings
- [ ] New unit tests: 9 `unescape_echo_*` tests + `parse_set_value_quoted_concat` test
- [ ] Updated `parse_set_value_with_spaces` test to reflect correct psql token-concatenation semantics

### DIFF
--- a/src/repl/mod.rs
+++ b/src/repl/mod.rs
@@ -7463,4 +7463,75 @@ mod tests {
         let result = super::unescape_echo("\\400");
         assert_eq!(result.as_bytes(), &[0x00]);
     }
+
+    // -- postgres_dba patterns (Copyright 2026) --------------------------------
+    //
+    // These tests replicate the exact \echo calls from
+    // https://github.com/NikolayS/postgres_dba/blob/master/start.psql
+    // to verify that split_params + unescape_echo together produce the
+    // correct ANSI output, matching what psql does.
+
+    #[test]
+    fn postgres_dba_menu_header_split_then_unescape() {
+        // start.psql line 2: \echo '\033[1;35mMenu:\033[0m'
+        // split_params strips the surrounding single quotes, then
+        // unescape_echo converts \033 to ESC (0x1b).
+        let raw = "'\\033[1;35mMenu:\\033[0m'";
+        let joined = crate::metacmd::split_params(raw).join(" ");
+        assert_eq!(joined, "\\033[1;35mMenu:\\033[0m");
+        let result = super::unescape_echo(&joined);
+        // First byte must be ESC (0x1b = 27).
+        assert_eq!(result.as_bytes()[0], 0x1b);
+        // Bold magenta on: [1;35m
+        assert!(result.contains("[1;35m"));
+        // Reset: ESC[0m
+        assert!(result.ends_with("\x1b[0m"));
+        // The literal text "Menu:" must be present.
+        assert!(result.contains("Menu:"));
+    }
+
+    #[test]
+    fn postgres_dba_error_banner_split_then_unescape() {
+        // start.psql line 219:
+        //   \echo '\033[1;31mError:\033[0m Unknown option! Try again.'
+        // split_params strips quotes; unescape_echo resolves \033.
+        let raw = "'\\033[1;31mError:\\033[0m Unknown option! Try again.'";
+        let joined = crate::metacmd::split_params(raw).join(" ");
+        assert_eq!(
+            joined,
+            "\\033[1;31mError:\\033[0m Unknown option! Try again."
+        );
+        let result = super::unescape_echo(&joined);
+        // First byte is ESC.
+        assert_eq!(result.as_bytes()[0], 0x1b);
+        // Bold red on: [1;31m
+        assert!(result.contains("[1;31m"));
+        // The literal error text must survive.
+        assert!(result.contains("Error:"));
+        assert!(result.contains("Unknown option! Try again."));
+        // Reset sequence present.
+        assert!(result.contains("\x1b[0m"));
+    }
+
+    #[test]
+    fn postgres_dba_plain_echo_no_escape() {
+        // start.psql line 79: \echo 'Bye!'
+        // No escape sequences; split_params strips quotes, output is literal.
+        let raw = "'Bye!'";
+        let joined = crate::metacmd::split_params(raw).join(" ");
+        let result = super::unescape_echo(&joined);
+        assert_eq!(result, "Bye!");
+    }
+
+    #[test]
+    fn postgres_dba_menu_item_echo_preserves_spacing() {
+        // start.psql line 3 (representative plain menu line):
+        //   \echo '   0 – Node and current database information'
+        // Spaces inside quotes must be preserved.
+        let raw = "'   0 \u{2013} Node and current database information'";
+        let joined = crate::metacmd::split_params(raw).join(" ");
+        let result = super::unescape_echo(&joined);
+        assert!(result.starts_with("   0"));
+        assert!(result.contains("Node and current database information"));
+    }
 }


NikolayS:
## REV Code Review — PR #596

**fix(repl): yolo auto-enables text2sql; fix \\echo escapes and \\set concat**

Reviewed by 5 parallel agents: security, bugs, tests, guidelines, docs.

---

### Summary table

| # | Agent | Finding | Severity | Confidence | Classification |
|---|-------|---------|----------|------------|----------------|
| 1 | bugs | `\set` concat breaks multi-word values | HIGH | 9 | **BLOCKING** |
| 2 | bugs | `\yolo` `SetInputMode` guard resets `exec_mode` to Interactive | MEDIUM | 8 | **BLOCKING** |
| 3 | bugs | `\echo` (not `\qecho`) is unchanged — asymmetric fix | MEDIUM | 8 | **BLOCKING** |
| 4 | tests | No test for `\set` concat / quote stripping behaviour | MEDIUM | 8 | **BLOCKING** |
| 5 | guidelines | Commit subject is 67 chars (limit 50) | LOW | 10 | NON-BLOCKING |
| 6 | docs | README example still shows SQL-mode prompt after `\yolo` (no text2sql prompt shown) | LOW | 9 | NON-BLOCKING |
| 7 | docs | README toggle note: dropped `\sql` toggle; only `\set TEXT2SQL_SHOW_SQL on` mentioned | LOW | 8 | NON-BLOCKING |
| 8 | security | No injection surface; octal fix is correct and safe | INFO | 10 | NON-BLOCKING |

---

### BLOCKING findings

#### [BUG-1] `\set` concat silently drops spaces in multi-word values — confidence 9, HIGH

The new `parse_set` logic calls `split_params(raw_value).concat()`.  `split_params` splits on whitespace and returns tokens; `.concat()` joins them with no separator.

```
\set X hello world
```

Before: `X = 'hello world'`  (trimmed tail of `splitn(2)`)
After:  `X = 'helloworld'`   (tokens joined without space)

This is a **regression** for every existing `\set VAR value with spaces` usage.  The correct fix is `.join(" ")` (matching the `\qecho` treatment in the same PR), not `.concat()`.

The psql-compatibility expectation is also clear from the existing test `parse_set_value_with_spaces` in the file, which asserts `"hello world"` (with space).

**Fix:** change `.concat()` to `.join(" ")` in `parse_set`.

---

#### [BUG-2] `SetInputMode` handler resets `exec_mode` to Interactive — confidence 8, MEDIUM

The PR makes `\yolo` dispatch `SetExecMode(Yolo)` and then — in the same handler — sets `input_mode = Text2Sql` via:

```rust
if mode == ExecMode::Yolo {
    settings.input_mode = InputMode::Text2Sql;
}
```

But `SetInputMode` is handled by the *other* branch, which always resets `exec_mode = Interactive`.  The new Yolo path uses `SetExecMode`, not `SetInputMode`, so the reset is not triggered here — that is intentional.

However: if the user later types `\t2s` (which dispatches `SetInputMode(Text2Sql)`), it resets `exec_mode` to Interactive, which is the existing behaviour and is correct.  But `\yolo` itself also emits a single `SetExecMode(Yolo)` and then patches `input_mode` inside that handler — it does **not** go through `SetInputMode`, so the "switch input mode → reset exec" safety valve is bypassed.  The asymmetry means `\mode` will show `input_mode=text2sql, exec_mode=yolo`, but arriving there via `\t2s` would have reset exec to interactive.

The real risk: the inline `settings.input_mode = InputMode::Text2Sql;` inside `SetExecMode` handler is a hidden side-effect that violates the single-responsibility of that match arm.  A cleaner design would emit a new composite `MetaResult::SetYoloMode` (sets both fields) or call a helper, so that all three `SetInputMode` / `SetExecMode` / `SetYoloMode` handlers each have one responsibility and the exec-reset logic in `SetInputMode` is not circumvented.

This is correctness-impacting if users expect `\t2s` after `\yolo` to be idempotent — it silently disables yolo.

**Fix:** introduce `MetaResult::SetYoloMode` that sets both fields atomically, or keep `YoloMode` as `SetExecMode(Yolo)` but remove the inline `input_mode` mutation and document that `\yolo` requires separate `\t2s` activation (contradicting the PR description).

---

#### [BUG-3] `MetaCmd::Echo` handler is unchanged — asymmetric — confidence 8, MEDIUM

The PR fixes `MetaCmd::QEcho` and `MetaCmd::Warn` to use `split_params` + `unescape_echo`, but `MetaCmd::Echo` at line 2494 is left as:

```rust
MetaCmd::Echo => {
    println!("{}", parsed.pattern.as_deref().unwrap_or(""));
```

psql's `\echo` and `\qecho` have identical quoting and escape semantics.  Leaving `\echo` without the fix means `\echo 'hello world'` still prints with the surrounding quotes rather than stripping them, while `\qecho 'hello world'` strips them.  This is a new behavioural divergence introduced by this PR.

**Fix:** apply the same `split_params` + `unescape_echo` treatment to `MetaCmd::Echo`.

---

#### [TEST-1] No regression test for the `\set` concat path — confidence 8, MEDIUM

The PR adds backslash-escaped-quote support to `split_params` and changes the `\set` value parsing, but adds no test covering:
- `\set X 'hello world'` (quoted value — should yield `hello world`)
- `\set X hello world` (multi-word unquoted — should yield `hello world`, not `helloworld`)
- `\set X 'it'\''s'` (escaped quote in value)

Given that BUG-1 exists, a test here would have caught it.

**Fix:** add unit tests for `parse_set` with quoted and multi-word values.

---

### NON-BLOCKING findings

#### [GUIDE-1] Commit subject exceeds 50-char limit — confidence 10, LOW

> `fix(repl): yolo auto-enables text2sql; fix \echo escapes and \set concat`

That is 67 characters.  CLAUDE.md / Conventional Commits style: subject < 50 chars.  Consider splitting into two commits or shortening to e.g. `fix(repl): yolo enables t2s; fix echo and set`.

---

#### [DOC-1] README yolo example does not show text2sql prompt — confidence 9, LOW

The PR says `\yolo` now enables text2sql, but the README example still shows:

```
postgres=# what is DB size?
```

which looks identical to the pre-existing text2sql example.  There is no visual cue in the example that text2sql was auto-activated by `\yolo`.  A clearer before/after example — starting from SQL mode and showing the mode transition message — would help users understand the new combined behaviour.

---

#### [DOC-2] README toggle note drops `\sql` / `\interactive` — confidence 8, LOW

The PR changes:
> `Toggle back with \sql or \interactive. Show/hide the SQL preview in text2sql mode with \set TEXT2SQL_SHOW_SQL off.`

to:
> `Toggle back with \sql or \interactive. Show/hide the SQL preview box with \set TEXT2SQL_SHOW_SQL on.`

The new text only mentions `\set TEXT2SQL_SHOW_SQL on` (to re-enable the box), but omits `off` (to hide it).  The previous text mentioned `off` specifically because that is the non-default action a user coming from `\yolo` mode would want.  Suggesting `on` only is confusing in context.

---

### Verdict

**Do not merge as-is.**

3 BLOCKING bugs + 1 BLOCKING missing test must be fixed first:
1. `\set` concat → use `.join(" ")` not `.concat()`
2. `MetaCmd::Echo` needs the same `split_params`/`unescape_echo` fix as `\qecho`
3. Consider the `SetExecMode` inline `input_mode` side-effect (atomicity / symmetry)
4. Add `parse_set` tests with quoted and multi-word values

The octal overflow fix and the `split_params` backslash-escape addition are correct and well-tested.

---
*BLOCKING: 4 | NON-BLOCKING: 3*
*Review generated by Claude Code (claude-sonnet-4-6) — 2026-03-17*

PR #609, opened 2026-03-18
fix(ai): DDL from /ask must prompt before executing

NikolayS:
## Summary

- `is_write_query()` only matched DML (`INSERT`/`UPDATE`/`DELETE`/`MERGE`), so DDL like `CREATE INDEX` was classified as read-only and **auto-executed without any confirmation prompt**
- Users asking `/ask What indexes should I add for my orders table?` had those indexes actually created on their database silently
- Fixes the bug where running `/ask` twice showed the AI saying "You've already added the indexes" (because they were actually created on the first run)

## Changes

Expanded `is_write_query()` to cover:
- DDL: `CREATE`, `DROP`, `ALTER`, `TRUNCATE`, `RENAME`
- Privilege control: `GRANT`, `REVOKE`
- Maintenance: `VACUUM`, `CLUSTER`, `REINDEX`, `REFRESH`

These now go through the `Execute (write query)? [Y/n/e]` prompt like DML does.

## Test plan
- [ ] Added unit tests for `CREATE INDEX`, `DROP TABLE`, `ALTER TABLE`, `TRUNCATE`, `GRANT`, `REVOKE`
- [ ] All 9 `write_query_*` tests pass

### DIFF
--- a/src/pager.rs
+++ b/src/pager.rs
@@ -126,6 +126,12 @@ impl Drop for TerminalGuard {
         // the terminal is always restored, including during panics.
         let _ = terminal::disable_raw_mode();
         let _ = execute!(io::stdout(), LeaveAlternateScreen);
+        // Reset the scroll region on the main screen buffer so the REPL can
+        // re-install its own DECSTBM constraint.  Without this the main buffer
+        // inherits whatever scroll region was set before the pager launched,
+        // leaving the cursor stuck at row 1.
+        let _ = io::stderr().write_all(b"\x1b[r");
+        let _ = io::stderr().flush();
     }
 }
 
--- a/src/repl/execute.rs
+++ b/src/repl/execute.rs
@@ -935,9 +935,10 @@ pub(super) async fn execute_query_interactive(
             sl.teardown_scroll_region();
         }
         run_pager_for_text(settings, &text, &captured);
-        // Re-establish scroll region and re-render status bar after pager exits.
+        // Re-establish scroll region, reposition cursor to bottom of scroll
+        // region, and re-render status bar after pager exits.
         if let Some(ref sl) = settings.statusline {
-            sl.setup_scroll_region();
+            sl.setup_scroll_region_and_restore_cursor();
             sl.render();
         }
     } else {
@@ -1054,9 +1055,10 @@ pub(super) async fn execute_query_extended_interactive(
             sl.teardown_scroll_region();
         }
         run_pager_for_text(settings, &text, &captured);
-        // Re-establish scroll region and re-render after pager exits.
+        // Re-establish scroll region, reposition cursor to bottom of scroll
+        // region, and re-render after pager exits.
         if let Some(ref sl) = settings.statusline {
-            sl.setup_scroll_region();
+            sl.setup_scroll_region_and_restore_cursor();
             sl.render();
         }
     } else {
--- a/src/statusline.rs
+++ b/src/statusline.rs
@@ -143,6 +143,25 @@ impl StatusLine {
         let _ = io::stderr().flush();
     }
 
+    /// Re-install the scroll region and reposition the cursor to the bottom
+    /// of the scroll region.
+    ///
+    /// Call this after returning from the built-in pager.  `LeaveAlternateScreen`
+    /// restores the main screen buffer but the DECSTBM scroll region constraint
+    /// may still be active; `setup_scroll_region()` re-establishes it, which
+    /// moves the cursor to row 1 as a side-effect.  This method then moves the
+    /// cursor back to the bottom of the scroll region so the next prompt
+    /// appears at the correct position.
+    pub fn setup_scroll_region_and_restore_cursor(&self) {
+        self.setup_scroll_region();
+        if !self.enabled {
+            return;
+        }
+        let bottom = self.term_rows.saturating_sub(1);
+        let _ = write!(io::stderr(), "\x1b[{bottom};1H");
+        let _ = io::stderr().flush();
+    }
+
     /// Restore the full scroll region.  Call at REPL exit.
     pub fn teardown_scroll_region(&self) {
         if !self.enabled {


NikolayS:
## REV Code Review Report

- **MR:** NikolayS/rpg#609 — fix(ai): DDL from /ask must prompt before executing
- **Author:** NikolayS
- **AI-Assisted:** Yes

| Pipeline |
|----------|
| ✅ All checks pass |

---

### BLOCKING ISSUES (1)

**MEDIUM** `src/repl/ai_commands.rs:1270` — CTE-prefixed destructive statements bypass `is_write_query`
> `is_write_query` checks only the first whitespace token. A query like `WITH x AS (SELECT 1) DELETE FROM users` starts with `WITH`, which is not in the keyword list, so it is classified as read-only and **auto-executed without confirmation**. This also affects `build_explain_sql`: a CTE-prefixed DML would run `EXPLAIN ANALYZE` without `BEGIN/ROLLBACK` wrapping, actually persisting the write.
>
> The existing test `write_query_with_cte_select_is_false` explicitly marks CTE queries as non-write, but `WITH ... DELETE/INSERT/UPDATE` is valid, destructive PostgreSQL syntax.
>
> **Fix:** After extracting the first token, if it is `WITH`, scan forward to find the trailing statement keyword (`SELECT`, `INSERT`, `UPDATE`, `DELETE`, `MERGE`) after all CTE bodies, and classify based on that keyword.

---

### POTENTIAL ISSUES (8)

**MEDIUM** `src/repl/ai_commands.rs:765` — YOLO mode still auto-executes DDL without confirmation *(confidence: 7/10)*
> The expanded `is_write_query` now catches DDL, but in YOLO mode all write queries (including `DROP TABLE`, `TRUNCATE`, `GRANT`) are auto-executed with only a stderr warning. Consider a harder block for destructive DDL even in YOLO.

**LOW** `src/repl/mod.rs:2160` — If `enable_raw_mode()` fails, code falls into raw event loop anyway *(confidence: 6/10)*
> `let raw_enabled = terminal::enable_raw_mode().is_ok();` — if this fails, the event loop still runs but keyboard events may not be delivered correctly; Ctrl+C may kill the process via OS signal instead.
> **Suggestion:** Fall back to the `read_line` path if raw mode cannot be enabled.

**LOW** `src/repl/mod.rs:2166` — Ctrl+D treated same as Ctrl+C (abort), psql treats it as accept-input *(confidence: 5/10)*
> `(KeyCode::Char('c' | 'd'), KeyModifiers::CONTROL) | (KeyCode::Esc, _) => { break true; }` — In psql, Ctrl+D at a `\prompt` signals EOF and accepts the typed input (or empty), it does not abort the script.

**LOW** `src/repl/ai_commands.rs:1270` — Unrealistic SQL in `write_query_ddl_is_true` test *(confidence: 6/10)*
> `"create index concurrently ..."` uses a literal `"..."` which would never appear in real input. Weakens test credibility.
> **Suggestion:** Replace with `"create index concurrently idx on t(id)"`.

**LOW** `src/metacmd.rs` — Missing test for mixed quoted/unquoted `\set` tokens *(confidence: 5/10)*
> `\set X 'hello' world` (quoted + unquoted mix) is not tested. Should concatenate to `"helloworld"`.

**LOW** `src/repl/mod.rs:1030` — No test for `prompt_interrupted` state transitions *(confidence: 5/10)*
> The `exec_lines` loop breaking on `prompt_interrupted = true` is untested.

**LOW** `src/repl/mod.rs:2132` — Doc comment only mentions Ctrl+C, not Ctrl+D or Esc *(confidence: 6/10)*
> `apply_prompt` doc says "When the user presses Ctrl+C" but code also handles Ctrl+D and Esc as interrupt triggers.

**INFO** `src/repl/mod.rs` — Dead code empty `if` block in non-interactive path *(confidence: 5/10)*
> `if !prompt_text.is_empty() { // Suppress prompt text... }` — the block is empty, the comment is misleading (suppression happens via absence of any `eprint!`, not via this guard).

---

### Summary

| Security | 1 | 1 | 0 |
| Bugs | 0 | 2 | 0 |
| Tests | 0 | 3 | 0 |
| Guidelines | 0 | 1 | 0 |
| Docs | 0 | 1 | 0 |

---

**Result: ❌ BLOCKED — fix the CTE bypass before merge.**

*REV-assisted review*

NikolayS:
## REV-assisted review — PR #609

**CI:** All checks pass (CodeQL, Analyze/actions, Analyze/ruby, Analyze/rust, all Build jobs, all Checks).
**Diff scope:** `src/repl/ai_commands.rs` + `src/repl/mod.rs`
**Latest commit:** 2026-03-17 22:13:01 -0700 (05:13:01 UTC)

---

### BLOCKING issues: 0

The previous BLOCKING issue (stale `is_write_query` doc comment) was fixed in the latest commit — the doc block now fully enumerates all covered keyword categories (DML/DDL/DCL/Maintenance) and broadens the description beyond EXPLAIN-wrapping. No remaining BLOCKING issues.

---

### POTENTIAL issues (non-blocking)

**[TESTS-1]** Maintenance keywords `VACUUM`, `CLUSTER`, `REINDEX`, `REFRESH` and `RENAME` have no unit tests
- **Severity:** LOW | **Confidence:** 8/10
- DDL (CREATE/DROP/ALTER/TRUNCATE) and DCL (GRANT/REVOKE) each have dedicated test functions. The maintenance group and `RENAME` are not tested.
- Suggestion: add `write_query_maintenance_*` and `write_query_ddl_rename` test functions.

**[TESTS-2]** `/ask` write-query non-execution path (`AskChoice::No`) not unit tested
- **Severity:** MEDIUM | **Confidence:** 7/10
- The new `else if !read_only { AskChoice::No }` branch replaces the old `ask_yne_prompt` call. No test exercises this path to verify write queries in `/ask` interactive mode now return `AskChoice::No` instead of prompting.

**[DOCS-1]** In-line comment `// (write query — not executed in /ask mode; use \\t2s to execute)` goes to stderr via `eprintln!`
- **Severity:** LOW | **Confidence:** 6/10
- This is correct behavior (informational message on stderr), but it is not documented in a public-facing doc or SPEC change. Minor.

---

### Positive changes

- `to_uppercase()` → `to_ascii_uppercase()`: avoids locale-dependent Unicode case folding. Correct improvement.
- CTE (`WITH`) treated as write query: prevents DML-bypass via CTE prefix.
- `/ask` no longer auto-executes write queries — safe default. Shows an advisory message via `eprintln!`.
- Comprehensive new keywords: DDL (CREATE/DROP/ALTER/TRUNCATE/RENAME), DCL (GRANT/REVOKE), maintenance (VACUUM/CLUSTER/REINDEX/REFRESH).

---

**Result: 0 BLOCKING — ready to merge once CI is confirmed green.**

*REV-assisted review*

PR #616, opened 2026-03-18
fix(pager): restore cursor position and scroll region after pager exit

### LINKED ISSUE #611: Pager exit leaves cursor at row 1 with previous command history visible
## Bug

After exiting from a long result set in the built-in pager (press `q`), the cursor ends up at the very first row of the terminal and the previous command history (not the paged content) is visible on screen.

## Steps to reproduce
1. Run a query that produces many rows (triggers pager)
2. Browse pages, then press `q` to exit

## Actual
- Cursor positioned at row 1 (top of terminal)
- Terminal shows history of previous commands, not a clean prompt

## Expected
- Cursor at bottom near the prompt
- Terminal shows a clean state ready for next input

## Notes
The pager uses `EnterAlternateScreen`/`LeaveAlternateScreen` (crossterm). The StatusLine also installs a scroll region (DECSTBM). The interaction between these two when the pager exits likely repositions the cursor incorrectly.

NikolayS:
## Summary

- `TerminalGuard::drop` in `src/pager.rs` now emits `\x1b[r` (reset scroll region) to stderr after `LeaveAlternateScreen`, so the main screen buffer is unconstrained when returning to the REPL
- `StatusLine` gains `setup_scroll_region_and_restore_cursor()` in `src/statusline.rs`: re-installs the DECSTBM scroll region and then moves the cursor to `term_rows - 1` so the next prompt appears at the correct position
- Both `run_pager_for_text` call sites in `src/repl/execute.rs` now call `setup_scroll_region_and_restore_cursor()` instead of bare `setup_scroll_region()`

## Root cause

After the built-in pager exits, `LeaveAlternateScreen` fires and returns to the main screen buffer. The REPL then calls `setup_scroll_region()` to re-install DECSTBM, but DECSTBM unconditionally moves the cursor to the home position (row 1, col 1) as a side-effect. Without a subsequent cursor-reposition, the next prompt appeared at row 1 with previous history visible instead of at the bottom.

## Test plan

- [ ] Run a query that triggers the built-in pager (result > terminal height)
- [ ] Press `q` to exit the pager
- [ ] Verify the prompt appears at the bottom of the terminal with status bar intact
- [ ] Verify previous query history is not visible above the prompt

Closes #611

### DIFF
--- a/src/repl/mod.rs
+++ b/src/repl/mod.rs
@@ -6288,7 +6288,9 @@ mod tests {
         // All CTEs treated as write to prevent CTE-prefixed DML bypass.
         assert!(is_write_query("with cte as (select 1) select * from cte"));
         assert!(is_write_query("WITH data AS (SELECT 1) DELETE FROM t"));
-        assert!(is_write_query("WITH data AS (SELECT 1) INSERT INTO t VALUES (1)"));
+        assert!(is_write_query(
+            "WITH data AS (SELECT 1) INSERT INTO t VALUES (1)"
+        ));
         assert!(is_write_query("WITH x AS (SELECT 1) UPDATE t SET a = 1"));
     }
 


NikolayS:
## REV Code Review Report

- **MR:** NikolayS/rpg#616 — fix(pager): restore cursor position and scroll region after pager exit
- **Author:** NikolayS
- **AI-Assisted:** Yes

| Pipeline |
|----------|
| ✅ All checks pass |

---

### BLOCKING ISSUES (2)

**MEDIUM** `src/repl/ai_commands.rs` — CTE-prefixed write queries bypass `is_write_query`
> `WITH ... DELETE/UPDATE/INSERT` starts with `WITH`, which is not in the match list, so it is classified as read-only and auto-executed without confirmation in yolo mode. Additionally `build_explain_sql` wraps by `BEGIN/ROLLBACK` only when `is_write_query` returns true — CTE-prefixed DML bypasses this, causing writes to persist during `EXPLAIN ANALYZE`.
>
> **Fix:** After extracting the first token, if it is `WITH`, scan forward to find the actual statement keyword after the CTE bodies (INSERT/UPDATE/DELETE/MERGE) and classify based on that.

**MEDIUM** `src/repl/mod.rs` — No test for `WITH ... DELETE/UPDATE/INSERT` against `is_write_query` *(confidence: 9/10)*
> The existing test `write_query_with_cte_select_is_false` only tests `WITH ... SELECT`. There is no test for CTE-prefixed write queries. Adding one (`assert!(is_write_query("WITH cte AS (SELECT 1) DELETE FROM t"))`) would currently fail, confirming the bypass.
>
> **Fix:** Add tests for CTE-prefixed DML patterns; they must pass for the CTE detection fix to be considered correct.

---

### NON-BLOCKING (0)

---

### POTENTIAL ISSUES (3)

**MEDIUM** `src/pager.rs:128` — `ESC[r` in `TerminalGuard::drop` sent to stderr; `LeaveAlternateScreen` sent to stdout *(confidence: 7/10)*
> When stdout is redirected (piped), `LeaveAlternateScreen` goes to the pipe while scroll region reset goes to the terminal. The status line uses stderr for its scroll region, so stderr is arguably correct here — but it should be documented with a comment explaining why.

**LOW** `src/pager.rs` — No integration test for `TerminalGuard::drop` escape sequence *(confidence: 7/10)*
> No test verifies that `ESC[r` is emitted on drop. Terminal escape sequences are hard to test, but a regression in cursor positioning would only be caught manually.

**LOW** `src/statusline.rs` — No test for `setup_scroll_region_and_restore_cursor` *(confidence: 6/10)*
> The new method has the same pattern as `on_resize` which also lacks a dedicated test.

---

### Summary

| Security | 1 | 1 | 0 |
| Bugs | 0 | 1 | 0 |
| Tests | 1 | 2 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |

---

**Result: ❌ BLOCKED — fix the CTE bypass before merge.**

*REV-assisted review*

PR #662, opened 2026-03-20
feat(repl): add Lua-based custom meta commands

### LINKED ISSUE #659: Ability to write custom meta commands, maybe in Lua

NikolayS:
## Summary

- Adds a Lua 5.4 runtime (via `mlua` crate, feature-gated as `--features lua`) so users can write custom backslash commands in `~/.config/rpg/commands/*.lua`
- New `\commands` meta-command lists all loaded custom Lua commands with descriptions
- Unknown backslash commands are matched against the registry before the "invalid command" error fires, so `\my_cmd arg1 arg2` Just Works once a script is installed
- Lua API: `rpg.register_command`, `rpg.query` (read-only tx), `rpg.print`, `rpg.pager`, `rpg.version`, `rpg.dbname`
- Two example scripts in `examples/commands/`: `bloat_check.lua` and `db_info.lua`
- When the `lua` feature is not compiled in, the registry is always empty and execution returns a graceful error message

## Architecture

- `src/lua_commands.rs` — `LuaRegistry` struct, command discovery, `rpg` module installation, handler invocation
- `src/metacmd.rs` — added `MetaCmd::ListCustomCommands` variant + `\commands` parser
- `src/repl/mod.rs` — `lua_registry` field on `ReplSettings`, loaded at session start; dispatch in `dispatch_meta` for both `ListCustomCommands` and fallback in the `Unknown` branch
- `rpg.query()` uses a thread-local client pointer + `block_in_place` / `Handle::block_on` for safe sync execution inside an async context

## Test plan

- [ ] `cargo test` — all 1643 tests pass
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [ ] `cargo fmt --check` — clean
- [ ] Manual: place a `.lua` script in `~/.config/rpg/commands/`, run rpg, invoke `\commands` and the custom command
- [ ] Verify `\commands` prints "no custom commands" when the directory is empty/absent
- [ ] Verify rpg starts cleanly without `--features lua` (empty registry, no crash)

Closes #659

### DIFF
--- a/src/lua_commands.rs
+++ b/src/lua_commands.rs
@@ -463,16 +463,25 @@ impl LuaRegistry {
         dbname: &str,
         client: &tokio_postgres::Client,
     ) -> Result<(), String> {
+        // RAII guard: clears the thread-local client pointer on return *and*
+        // on panic, preventing a dangling pointer if `invoke` unwinds.
+        struct ClientPtrGuard;
+        impl Drop for ClientPtrGuard {
+            fn drop(&mut self) {
+                lua_impl::clear_client_ptr();
+            }
+        }
+
         lua_impl::update_dbname(&self.lua, dbname).map_err(|e| format!("rpg: lua: {e}"))?;
 
-        // Safety: client lives for the duration of the `block_in_place` call
-        // in `dispatch_meta`.  We clear the pointer before returning.
+        // Safety: `client` lives for the duration of the `block_in_place`
+        // call in `dispatch_meta`.  `ClientPtrGuard` ensures the pointer is
+        // cleared on both normal return and panic.
         unsafe {
             lua_impl::set_client_ptr(std::ptr::from_ref(client));
         }
-        let result = lua_impl::invoke(&self.lua, name, args);
-        lua_impl::clear_client_ptr();
-        result
+        let _guard = ClientPtrGuard;
+        lua_impl::invoke(&self.lua, name, args)
     }
 
     #[cfg(not(feature = "lua"))]


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/rpg#662 — feat(repl): add Lua-based custom meta commands
- **Author:** NikolayS
- **AI-Assisted:** Yes (Claude Code)
- **CI Status:** ⏳ pending (CodeQL analysis running)

---

### BLOCKING ISSUES (1)

**MEDIUM** `src/lua_commands.rs:500-506` — Client pointer not cleared on panic

> ```rust
> unsafe { lua_impl::set_client_ptr(std::ptr::from_ref(client)); }
> let result = lua_impl::invoke(&self.lua, name, args);
> lua_impl::clear_client_ptr();  // NOT called if invoke panics
> ```
> If `lua_impl::invoke` panics (rather than returning `Err`), `clear_client_ptr()` is never called, leaving a dangling raw pointer in the thread-local for the next invocation.
>
> **Fix:** Use a RAII guard:
> ```rust
> struct ClientPtrGuard;
> impl Drop for ClientPtrGuard {
>     fn drop(&mut self) { lua_impl::clear_client_ptr(); }
> }
> unsafe { lua_impl::set_client_ptr(std::ptr::from_ref(client)); }
> let _guard = ClientPtrGuard;
> lua_impl::invoke(&self.lua, name, args)
> ```

---

### NON-BLOCKING (6)

**MEDIUM** `src/repl/mod.rs:1229` — `LuaRegistry::load("")` in `Default::default()` scans the filesystem

> Every `ReplSettings::default()` call (including in 1600+ unit tests) scans `~/.config/rpg/commands/`. This introduces test non-determinism if a developer has `.lua` files installed, and doubles the I/O (Default scans with `""`, then `run_repl` scans again with the real dbname).
>
> **Suggestion:** Add `LuaRegistry::empty()` returning an empty registry without disk access, use it in `Default`, and keep `LuaRegistry::load()` only in `run_repl`.

**MEDIUM** `src/lua_commands.rs:207-217` — Silent command name collision

> If two `.lua` scripts register the same command name, `commands` gets both entries but `rpg._handlers[name]` is overwritten by the second. `get()` returns metadata from the first script, but `execute_command` runs the handler from the second. The mismatch is silent.
>
> **Suggestion:** On duplicate name, print a warning and skip the second registration.

**MEDIUM** `src/lua_commands.rs:143` — `ALL_SAFE` Lua stdlib includes `io` library

> `mlua::StdLib::ALL_SAFE` includes Lua's `io` module, so any `.lua` script in `~/.config/rpg/commands/` can read arbitrary files (e.g. `~/.pgpass`, SSH keys). Scripts are user-authored but this is a wider attack surface than necessary.
>
> **Suggestion:** Restrict to `STRING | TABLE | MATH | PACKAGE` (drop `io` and `os.execute`). Document the restricted sandbox in the module doc.

**MEDIUM** `examples/commands/bloat_check.lua:1-5` — Misleading comment vs. SQL

> The comment promises `\bloat_check myschema.%` (LIKE pattern), but the SQL uses `where schemaname = '%s'` (equality). A user following the example will be confused when `%` is treated literally.
>
> **Suggestion:** Either change the SQL to `where schemaname like '%s'`, or remove the pattern example from the comment.

**LOW** `src/repl/mod.rs:~3238` — `block_in_place` fragility on single-thread runtime

> `block_in_place` panics on a `current_thread` runtime. The REPL always uses `new_multi_thread()` so this is safe today, but integration tests or future callers using a single-thread runtime would panic even if the Lua handler never calls `rpg.query()`.
>
> **Suggestion:** Add a doc comment on `execute_command`: _"Must be called from within `tokio::task::block_in_place` on a multi-thread Tokio runtime."_

**LOW** `src/lua_commands.rs` — No unit tests for `LuaRegistry`

> The Lua loading path (scan dir, parse scripts, register commands) has zero test coverage. Key untested scenarios: absent directory, syntax error in script, duplicate name, missing `handler` field.
>
> **Suggestion:** Add `#[cfg(test)]` tests using `tempfile` to exercise these paths without touching the real `~/.config/rpg/commands/` directory.

---

### POTENTIAL ISSUES (2)

**MEDIUM** `src/lua_commands.rs:108-120` — Unsafe pointer safety relies on external invariants *(confidence: 6/10)*

> The thread-local raw pointer safety depends on `dispatch_meta` always wrapping the call in `block_in_place`. If future refactoring removes or bypasses that wrapper, the unsafe code becomes unsound. A re-entrancy check (assert the pointer is null before setting it) would make the invariant self-enforcing.

**LOW** `examples/commands/bloat_check.lua:35` — String-formatted schema name in SQL *(confidence: 5/10)*

> `string.format("... where schemaname = '%s' ...", schema)` where `schema = args[1]`. A schema name with a single quote causes a syntax error (read-only tx means no data impact, but it's a bad pattern in example code). Consider `schema:gsub("'", "''")` or a note explaining the limitation.

---

### Summary

|------|----------|-----------|---------|
| CI/Pipeline | 0 (pending) | 0 | 0 |
| Security | 0 | 1 | 1 |
| Bugs | 1 | 1 | 0 |
| Tests | 1 | 0 | 0 |
| Guidelines | 1 | 0 | 0 |
| Docs | 1 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

---

### SOC2 COMPLIANCE (2)

**HIGH** SOC2: Code Review — No assigned reviewer

> MR has no assigned reviewer (author is NikolayS).
> **Requirement:** CC6.1 — Logical Access Controls
> **Action:** Assign a reviewer other than the author before merge.

**HIGH** SOC2: Linked Issue — Issue reference present ✅

> Description contains `Closes #659` — compliant.

*(SOC2 findings are informational for this project — not blocking per project policy.)*

---
*REV-assisted review (parallel AI analysis)*

PR #725, opened 2026-03-24
fix(connection): check socket file existence in default_host(), not directory

### LINKED ISSUE #721: fix(connection): default_host() checks directory existence instead of socket file
## Problem

`default_host()` in `src/connection.rs` (~line 365) checks whether the candidate socket *directory* exists (`is_dir()`), not whether a socket *file* exists inside it.

On this machine `/var/run/postgresql` exists (contains `13-main.pg_stat_tmp`) but has no `.s.PGSQL.5432` socket. rpg picks it as the default host and then fails with:

```
rpg: connection to server at "/var/run/postgresql", port 5432 failed: No such file or directory (os error 2)
```

psql handles this correctly — it checks for the actual socket file before selecting a default.

## Fix

Replace the `is_dir()` check with a check for the socket file:

```rust
// Instead of:
PathBuf::from(dir).is_dir()

// Check for the actual socket file:
PathBuf::from(dir).join(format!(".s.PGSQL.{}", port)).exists()
```

The port is needed for the socket filename. If port is not yet resolved at this point, fall through to the next candidate rather than selecting a directory that has no active socket.

## Discovered

Full connection matrix audit, 2026-03-24 (Section B).

NikolayS:
Closes #721

Previously checked if the candidate socket *directory* existed. Now checks if the actual socket file (.s.PGSQL.<port>) exists inside it, preventing false positives when the directory exists but no Postgres is running.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -157,6 +157,18 @@ jobs:
           --health-timeout 5s
           --health-retries 15
           --health-start-period 5s
+      postgres-scram:
+        image: postgres:16
+        env:
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 15433:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 2s
+          --health-timeout 5s
+          --health-retries 15
+          --health-start-period 5s
     env:
       TEST_PGHOST: localhost
       TEST_PGPORT: "15432"
@@ -199,6 +211,11 @@ jobs:
           TEST_PG_TLS_HOST: localhost
           TEST_PG_TLS_PORT: "5433"
           TEST_PG_TLS_PASSWORD: testpass
+          TEST_PG_SCRAM_HOST: localhost
+          TEST_PG_SCRAM_PORT: "15433"
+          TEST_PG_SCRAM_USER: postgres
+          TEST_PG_SCRAM_PASSWORD: postgres
+          TEST_PG_SCRAM_DATABASE: postgres
         run: bash tests/compat/test-connections.sh target/release/rpg
 
   coverage:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3459,9 +3459,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.9"
+version = "0.103.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7df23109aa6c1567d1c575b9952556388da57401e4ace1d15f79eedad0d8f53"
+checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
 dependencies = [
  "aws-lc-rs",
  "ring",
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -2046,15 +2046,21 @@ async fn connect_one(
             Err(e) => return Err(e),
         },
 
-        SslMode::Prefer => match connect_tls_default(pg_config, params).await {
-            Ok((c, info)) => (c, Some(info)),
-            Err(_) => {
-                // sslmode=prefer: silently fall back to a plain connection
-                // when TLS is unavailable. This matches psql's default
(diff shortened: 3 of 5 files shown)


NikolayS:
## REV: fix(connection): check socket file existence in default_host(), not directory

**Verdict: REQUEST CHANGES** — the fix is directionally correct and the bug is real, but there are two issues worth addressing before merge.

---

### 1. Correctness: directory scan matches `.s.PGSQL.<port>.lock` too

The `.starts_with(".s.PGSQL.")` predicate matches both the Unix-domain socket (`.s.PGSQL.5432`) and the lock file (`.s.PGSQL.5432.lock`). The lock file is a regular file created alongside the socket by postmaster. In practice this means `has_socket` can be `true` if only the lock file is present — which could happen in a crash scenario where the socket was cleaned up but the lock file wasn't. Not a common case, but it's a latent false positive that partially reintroduces the bug this PR is trying to fix.

The fix is straightforward: check file type explicitly.

```rust
e.file_type().map(|ft| ft.is_socket()).unwrap_or(false)
    && e.file_name().to_string_lossy().starts_with(".s.PGSQL.")
```

Or alternatively, since the port is already known (defaults to 5432), skip the directory scan entirely and check directly:

```rust
let socket = path.join(format!(".s.PGSQL.{}", DEFAULT_PORT));
if socket.exists() {
    return (*dir).to_owned();
}
```

The direct check is O(1) vs O(n) directory scan and cannot accidentally match a socket for a different port (see point 2).

---

### 2. Correctness: scanning all sockets ignores the target port

The current implementation returns a directory if any `.s.PGSQL.*` socket exists in it, regardless of port. If `/tmp/.s.PGSQL.5433` exists but there is no `.s.PGSQL.5432`, the code returns `/tmp` and then the connection attempt to port 5432 fails. The user gets a confusing error: the directory was "detected" as having Postgres, but the connection still fails.

libpq sidesteps this by not probing at all — it just tries to connect and lets the OS return ENOENT or ECONNREFUSED. If we want to probe, we should probe the specific socket that will actually be used.

---

### 3. Performance: `read_dir()` on every default connection attempt

`default_host()` is called from `Default for ConnParams` and from `resolve_params()`, which is on every connection path when no host is specified. `read_dir()` is a syscall + directory traversal. For the common case (a single Postgres instance), this is one `stat()` call if we use the direct path check instead of `read_dir()`. Not a significant concern in practice for a terminal tool, but the direct approach is both simpler and faster.

---

### 4. Error handling: silent `.ok()` on `read_dir()`

Silently ignoring `read_dir()` errors is acceptable here — if we can't read the directory, we can't tell if a socket exists, and falling back to `localhost` is the right default behavior. No concern.

---

### 5. Doc comment: backtick-quoted `PostgreSQL` is unconventional

`contains an active \`PostgreSQL\` socket file` — the backticks around `PostgreSQL` look odd; that's identifier formatting, not prose formatting. Just write it as plain text: "contains an active PostgreSQL socket file".

---

### 6. Unrelated: version bump and CHANGELOG belong in PR #720

This PR contains two commits: the fix (correct) and a `chore: bump version to 0.8.1` commit. The `Cargo.toml` bump to 0.8.1 and the new `## [0.8.1]` CHANGELOG entry were already added in the separate version-bump PR #720. Having both in this PR creates a merge conflict risk and makes the history harder to read. The version bump commit should be dropped from this PR.

Note also: the CHANGELOG entry for 0.8.1 describes fixes #708 and #711, not this fix (#721). Either this fix should be added to that entry, or the entry is misplaced here.

---

### Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | Lock file matches prefix, partial false-positive reintroduced | Medium |
| 2 | Scan matches wrong-port sockets | Medium |
| 3 | `read_dir()` more expensive than needed | Low |
| 4 | Silent `read_dir()` error suppression | OK |
| 5 | Doc comment backtick style | Nit |
| 6 | Version bump / CHANGELOG leak from PR #720 | Should fix |

The simplest fix for 1, 2, and 3 together: replace the `read_dir()` scan with a single `path.join(".s.PGSQL.5432").exists()` check (using the resolved port, or the default if not yet resolved).

Issue #724, opened 2026-03-24
feat(connection): support comma-separated multi-host for failover (psql parity)


NikolayS:
## Problem

psql supports comma-separated hostnames for automatic failover:

```bash
psql -h "host1,host2,host3" -p 5432 -U postgres -c "SELECT 1"
```

rpg treats the entire comma-separated string as a single hostname, sends it to DNS, and fails.

## Expected

rpg should parse comma-separated hosts and try each in sequence, connecting to the first that succeeds. This is standard libpq behavior and is commonly used for HA setups with a primary and one or more standbys.

## psql docs reference

> If host is a comma-separated list of host names, each will be tried in the given order.

## Acceptance criteria

- [ ] `-h "host1,host2"` tries host1 first, falls back to host2
- [ ] `PGHOST=host1,host2` same behavior  
- [ ] `host=host1,host2` in connstring same behavior
- [ ] `postgres://host1,host2/db` URI form same behavior
- [ ] Stops at first successful connection
- [ ] Error message lists all tried hosts if all fail

## Discovered

Full connection matrix audit, 2026-03-24 (Section F, F5).

Issue #723, opened 2026-03-24
fix(connection): PGCONNECT_TIMEOUT actual elapsed ~2x the configured value


NikolayS:
## Problem

Setting `PGCONNECT_TIMEOUT=1` results in an actual timeout of ~2 seconds.

```bash
time PGCONNECT_TIMEOUT=1 rpg -h 10.255.255.1 -U postgres -c "SELECT 1" 2>&1
# => real 0m2.001s  (expected ~1s)
```

## Likely cause

Possible unit mismatch (ms vs s) or the timeout is applied per-attempt and there are 2 attempts (IPv4 + IPv6, or primary + fallback). Check how `PGCONNECT_TIMEOUT` is passed to tokio's connection timeout.

## Expected

`PGCONNECT_TIMEOUT=1` should time out in approximately 1 second, matching psql behavior.

## Discovered

Full connection matrix audit, 2026-03-24 (Section E, E9).

Issue #709, opened 2026-03-22
test: verify all connection paths work correctly


NikolayS:
## Motivation

@mlusetti reported on X (2026-03-22) that rpg gives `connection failed: db error` where psql connects fine. The error message issue is fixed in #708, but the underlying connection compatibility needs a full audit.

### A. TCP connections

| # | Method | Example | Expected |
|---|--------|---------|---------|
| A1 | Basic TCP | `rpg -h localhost -p 5432 -U postgres` | ✅/❌ |
| A2 | TCP with password | `rpg -h host -p 5432 -U user -W` | ✅/❌ |
| A3 | TCP with PGPASSWORD | `PGPASSWORD=x rpg -h host -U user` | ✅/❌ |
| A4 | TCP with .pgpass | `~/.pgpass` file, no password prompt | ✅/❌ |
| A5 | TCP no port (default 5432) | `rpg -h localhost -U postgres` | ✅/❌ |
| A6 | TCP with database | `rpg -h localhost -U u -d mydb` | ✅/❌ |

### B. Unix socket connections

| # | Method | Example | Expected |
|---|--------|---------|---------|
| B1 | Default socket | `rpg -U postgres` (no -h) | ✅/❌ |
| B2 | Explicit socket path | `rpg -h /var/run/postgresql -U postgres` | ✅/❌ |
| B3 | Socket with port suffix | `rpg -h /var/run/postgresql -p 5432 -U postgres` | ✅/❌ |
| B4 | PGHOST as socket dir | `PGHOST=/tmp rpg -U postgres` | ✅/❌ |

### C. URI / connection string

| # | Method | Example | Expected |
|---|--------|---------|---------|
| C1 | postgres:// URI | `rpg postgres://user:pass@host/db` | ✅/❌ |
| C2 | postgres:// no password | `rpg postgres://user@host/db` | ✅/❌ |
| C3 | postgresql:// scheme | `rpg postgresql://user@host/db` | ✅/❌ |
| C4 | URI with sslmode | `rpg postgres://user@host/db?sslmode=require` | ✅/❌ |
| C5 | URI with socket | `rpg postgres:///db?host=/var/run/postgresql` | ✅/❌ |
| C6 | key=value connstring | `rpg "host=localhost user=u dbname=d"` | ✅/❌ |

### D. SSL modes

| # | sslmode | Expected |
|---|---------|---------|
| D1 | disable | ✅/❌ |
| D2 | prefer (default) | ✅/❌ |
[...]

PR #733, opened 2026-03-25
test(connection): fill #709 matrix gaps — B3/B4, C6-C8, E6-E9, F4-F7, G3/G4

NikolayS:
## What

Adds 15 new integration tests to `tests/connection_paths.rs` covering cases from the #709 audit matrix that were not previously in CI.

## New tests

| Test | Group | What it checks |
|------|-------|----------------|
| B3 | Socket error paths | No socket in dir — clean error, no panic |
| B4 | Socket error paths | Socket exists, wrong user — auth error, no panic |
| C6 | URI edge cases | Non-numeric port in URI — parse error |
| C7 | URI edge cases | `?port=` query param regression guard (C5 fix, PR #731) |
| C8 | URI edge cases | `?host=` query param ignores authority host |
| E6 | Env vars | `PGSSLMODE=disable` suppresses TLS (ssl=f in pg_stat_ssl) |
| E7 | Env vars | `PGDATABASE` from env, no `-d` flag |
| E8 | Env vars | `PGPASSFILE` reads .pgpass file |
| E9 | Env vars | `PGCONNECT_TIMEOUT` — elapsed documented; issue #723 noted inline |
| F4 | Error quality | Error includes database name |
| F5 | Error quality | Error includes user name |
| F6 | Error quality | Error includes host on connection refused |
| F7 | Error quality | Multi-host comma-separated — canary for #724; no-panic check |
| G3 | Auth | Wrong SCRAM password — clear error, no panic |
| G4 | Auth | `PGPASSWORD` takes precedence over `PGPASSFILE` |

## Notes

- F7 is a **canary test for issue #724** (multi-host not yet implemented). It currently only checks for no-panic; once #724 is fixed the assertion should be tightened to `assert_eq!(code, 0)`.
- E9 documents the known issue #723 (timeout is ~2×configured). The bound is loose until #723 is fixed.
- D5/D6 (verify-ca/verify-full) remain SKIPped pending issue #712.
- B4, C4 (socket tests) skip cleanly when no host-mode PG socket is present in CI.

Closes part of #709.

### DIFF
--- a/tests/connection_paths.rs
+++ b/tests/connection_paths.rs
@@ -1186,15 +1186,14 @@ fn e8_pgpassfile_env() {
     let dir = tempfile::tempdir().expect("e8: failed to create tempdir");
     let pgpass = dir.path().join(".pgpass");
 
-    // Wildcard entry so it matches any database/user on the trust instance.
-    // Trust auth ignores the password, so any value works.
+    // Write the correct password (from env or default) to .pgpass.
+    // The point of the test is that PGPASSFILE is read; we need a valid
+    // password here so password-auth servers (SCRAM in CI) also succeed.
     let host = trust_host();
     let port = trust_port();
-    std::fs::write(
-        &pgpass,
-        format!("{host}:{port}:*:*:pgpassword_test_value\n"),
-    )
-    .expect("e8: failed to write .pgpass");
+    let password = trust_password().unwrap_or_default();
+    std::fs::write(&pgpass, format!("{host}:{port}:*:*:{password}\n"))
+        .expect("e8: failed to write .pgpass");
 
     // chmod 600 — libpq ignores .pgpass files that are world-readable
     #[cfg(unix)]


NikolayS:
## REV Code Review — PR #733: connection path matrix gap tests (B3/B4, C6-C8, E6-E9, F4-F7, G3/G4)

Reviewed diff: `origin/main..HEAD -- tests/connection_paths.rs` (525 lines, 15 new tests)

---

### MAJOR — E9: final `assert!` is a tautology — timeout message never verified

**Location:** `tests/connection_paths.rs` (E9 block, final `assert!`, diff lines ~316-322)

```rust
assert!(
    combined.contains("timeout")
        || combined.contains("timed out")
        || combined.contains("connection refused")
        || code != 0,          // ← always true; already asserted above
    "e9: expected timeout error\n..."
);
```

The comment says "Verify a timeout-related error appears", but the trailing `|| code != 0` makes the entire condition vacuously true — `code != 0` was already enforced by the preceding `assert_ne!(code, 0, ...)`. The message-content check is completely unenforced. A test that silently passes even when rpg emits no recognisable timeout message defeats the purpose of the assertion.

**Suggestion:** Drop the `|| code != 0` arm:
```rust
assert!(
    combined.contains("timeout")
        || combined.contains("timed out")
        || combined.contains("connection refused"),
    "e9: expected timeout error message\nstdout: {stdout}\nstderr: {stderr}"
);
```

---

### MINOR — E9: `10.255.255.1` is not guaranteed unreachable in cloud CI

**Location:** `tests/connection_paths.rs` (E9, diff ~line 294)

```rust
cmd.args(["-h", "10.255.255.1", "-U", "nobody", "-w", "-c", "SELECT 1"])
```

`10.255.255.1` is a private address that _usually_ black-holes on GitHub Actions runners, but it is not standards-reserved for documentation/test purposes. Depending on the runner's VPC routing, it may route to an actual host or return `ECONNREFUSED` immediately rather than timing out, making the elapsed-time bound meaningless.

**Suggestion:** Use `192.0.2.1` (TEST-NET-1, RFC 5737), which is IANA-reserved and guaranteed non-routable in all compliant networks. This is the standard black-hole address for exactly this use case.

---

### MINOR — E6: `stdout.contains('f')` is an unreliable boolean check

**Location:** `tests/connection_paths.rs` (E6, diff ~lines 201-204)

```rust
assert!(
    stdout.contains('f'),
    "e6: expected ssl=f with PGSSLMODE=disable\nstdout: {stdout}"
);
```

The character `'f'` appears in a large number of strings — "failed", "refresh", "info", "default", etc. If rpg ever emits even a single diagnostic line containing 'f', this assertion passes regardless of the actual `pg_stat_ssl.ssl` value. Contrast with the `'1'` checks elsewhere in the suite, where `'1'` appearing in `SELECT 1` output is much more constrained.

**Suggestion:** Assert a more specific pattern that won't match incidental output, e.g.:
```rust
assert!(
    stdout.contains(" f\n") || stdout.contains(" f "),
    "e6: expected ssl=f with PGSSLMODE=disable\nstdout: {stdout}"
);
```
Or query `ssl::text` and assert `stdout.contains("false")`.

---

### MINOR — F7: "canary" test cannot detect when #724 is fixed

**Location:** `tests/connection_paths.rs` (F7, diff ~lines 435-443)

```rust
// Once #724 is fixed, this should be assert_eq!(code, 0, ...).
// For now we just verify it exits with a clear error (no panic).
let (stdout, stderr, code) = run(cmd);
let combined = format!("{stdout}{stderr}").to_lowercase();
assert!(
    !combined.contains("unwrap") && !combined.contains("panic"),
    ...
);
let _ = code; // expected to be nonzero until #724 is fixed
```

The comment says this test will "break once support lands", implying it will alert the team to tighten the assertion. However, the only assertion — no `unwrap`/`panic` in output — will also pass after #724 is fixed (a successful connection won't panic either). The test will silently continue passing in its loose form forever; nobody will be prompted to tighten it.

**Suggestion:** Assert `code != 0` explicitly to document the currently-broken state. When #724 lands and multi-host starts working, `code` will flip to 0 and this test will fail, forcing the tighten-up:
```rust
assert_ne!(
    code, 0,
    "f7: multi-host not yet supported (#724); expected failure\nstdout: {stdout}\nstderr: {stderr}"
);
```

---

### MINOR — B4: `CONN_SOCKET_PORT` / `CONN_SOCKET_DIR` not documented in module header

**Location:** `tests/connection_paths.rs` (B4, diff ~lines 55-57); module header (file lines 19-26)

```rust
let socket_port = std::env::var("CONN_SOCKET_PORT").unwrap_or_else(|_| "5437".to_owned());
let socket_dir  = std::env::var("CONN_SOCKET_DIR").unwrap_or_else(|_| "/var/run/postgresql".to_owned());
```

Every other configurable env var in the file has a corresponding helper function and is listed in the module-level documentation table (lines 19-26). `CONN_SOCKET_PORT` and `CONN_SOCKET_DIR` are the only two that are read inline and absent from the table, making them invisible to anyone who consults the header for CI configuration instructions.

**Suggestion:** Add them to the module doc table and optionally extract helper functions `socket_port()` / `socket_dir()` for consistency.

---

### NIT — C6: `#[ignore]` annotation is misleading

**Location:** `tests/connection_paths.rs` (C6, diff ~line 95)

```rust
#[ignore = "requires live Postgres — run via connection-tests CI job"]
fn c6_uri_bad_port() {
```

`c6_uri_bad_port` passes `postgres://localhost:notaport/postgres` which should be rejected at URI parse time, before any network I/O. A live Postgres is not required. The annotation is technically incorrect and may confuse contributors who want to run a quick subset of non-infrastructure tests locally.

**Suggestion:** Either remove `#[ignore]` entirely (letting it run in `cargo test`), or change the annotation text to something like `"parse-time test — safe to run locally without Postgres"`.

---

### NIT — E8, G4: `.unwrap()` without diagnostic message on path conversion

**Location:** diff ~lines 268, 514

```rust
.env("PGPASSFILE", pgpass.to_str().unwrap())
```

All other `.expect(...)` calls in the new tests include a descriptive message. This bare `.unwrap()` will produce an opaque panic if the temp path ever contains non-UTF-8 bytes.

**Suggestion:**
```rust
.env("PGPASSFILE", pgpass.to_str().expect("e8: pgpass path not UTF-8"))
```

---

### PRAISE — Several things done especially well

- **All 15 tests carry `#[ignore]`** — zero risk of breaking the standard `cargo test` run; only runs when the CI job explicitly passes `--include-ignored`.

- **Early-return skip pattern** (B4, E6) — using `eprintln!("SKIP …"); return;` keeps CI green when optional services are absent, rather than calling `assert_ignored!` macros or silently succeeding.

- **`env_remove("PGPASSWORD")` in E8** — actively purging ambient env before testing PGPASSFILE prevents silent false-positives when `PGPASSWORD` happens to be set in the test runner's environment.

- **chmod 600 on .pgpass files** (E8, G4) — correctly handles the libpq requirement that password files not be world-readable; the `#[cfg(unix)]` guard keeps the code portable.

- **Inline issue references** (#723 in E9, #724 in F7) — the reader can go straight to the issue for context on why a bound is deliberately loose. Good practice for intentional workarounds.

- **`tempfile::tempdir()`** in B3, E8, G4 — automatic cleanup on drop; no leftover artifacts even on test failure.

---

### Verdict: REQUEST_CHANGES

One MAJOR finding (E9 tautology renders the timeout message assertion inert) and the F7 canary issue should be fixed before merge. The remaining findings are minor/nit but worth addressing to keep the test suite trustworthy as the matrix fills in further.

Issue #578, opened 2026-03-17
(postponed) Fix --update: GitHub API returns 404


NikolayS:
\`rpg --update\` fails with:

```
rpg: update check failed: GitHub API returned status 404 Not Found
```

Likely caused by the repo URL being hardcoded to the old project-alpha path. Needs updating after repo rename to rpg.

Also: the Homebrew tap and install script are not yet available. These should only be documented once they work.

PR #735, opened 2026-03-25
fix(update): correct GitHub API URL + switch russh to ring backend

NikolayS:
Two independent bug fixes in one PR (both tiny, both in different files).

---

## fix(update): `--update` returns 404 (#578)

`GITHUB_API_URL` in `src/update.rs` was hardcoded to `NikolayS/project-alpha`. Every `--update` call since the repo rename has returned 404.

One-line fix: update to `NikolayS/rpg`.

---

### Problem

`russh = "0.57"` with default features pulls in `aws-lc-sys` — 67 MiB of BoringSSL C/C++ compiled from scratch via cmake. This adds minutes to every clean build and requires cmake as a system dependency.

### Fix

`russh = { version = "0.57", default-features = false, features = ["ring", "flate2"] }`

- `ring` crypto backend — already in the tree via `rustls-webpki`, no new C build
- `flate2` — retained for SSH compression support
- `rsa` removed — rpg uses key-based auth via `decode_secret_key`, which works with ring

### Verification

Before: `cargo tree -i aws-lc-sys` shows russh in the chain
After: russh no longer appears — `aws-lc-sys` remains only via the rustls→reqwest path (unavoidable)

1702 unit tests pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,20 +175,19 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.16.1"
+version = "1.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94bffc006df10ac2a68c83692d734a465f8ee6c5b384d8545a636f81d858f4bf"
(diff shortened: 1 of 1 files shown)


NikolayS:
Fixed: Cargo.lock regenerated and committed (commit 86ef936). The lock now removes aws-lc-rs from russh's dependency subtree — verified with `git diff Cargo.lock | grep aws-lc`.

Note on repository field / CONTRIBUTING.md: those reference the public repo name correctly (NikolayS/rpg). No change needed there.

PR #751, opened 2026-03-25
feat(connection): multi-host failover

### LINKED ISSUE #743: feat(connection): multi-host failover (comma-separated hosts, psql parity)
## Problem

psql supports comma-separated host lists for failover:
```
psql -h host1,host2,host3 -p 5432 -U postgres
```
rpg does not. This is a psql parity gap that affects HA setups (Patroni, pgBouncer, etc.).

## Spec

Follow libpq semantics exactly:
- `-h host1,host2,host3` — try each in order, use first that accepts connection
- `-p 5432,5433` — per-host ports (must have 1 port or N ports matching N hosts)
- URI form: `postgres://host1:5432,host2:5432/db`
- Connection string form: `host=host1,host2 port=5432`
- Stop at first successful connection
- Error message when all hosts fail: list which hosts were tried and why each failed

## Reference

- libpq docs: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-MULTIPLE-HOSTS
- tokio-postgres already parses multi-host connstrings via `Config::from_str` — leverage this

## Implementation hints

- `src/connection.rs`: `ConnParams` currently stores single host/port — extend to `Vec<(host, port)>`
- `connect()` function needs retry loop over hosts
- Error reporting: collect per-host errors, report all if all fail
- Integration tests: at least 2 Docker containers, test failover (first down, second up)

## Acceptance criteria

- `rpg -h host1,host2 -U u -d db` works (first host down → falls through to second)
- URI and connstring forms work
- Error lists all attempted hosts when all fail
- Integration tests cover: first-host-fails, all-fail, single-host (regression)
- Existing connection tests still pass

NikolayS:
Closes #743

Implements comma-separated multi-host connection with failover, matching psql/libpq semantics.

- `rpg -h host1,host2 -p 5432` tries hosts in order, uses first that accepts connection
- Per-host ports supported: `-h host1,host2 -p 5432,5433`
- Port reuse when fewer ports than hosts: last port is reused (matching libpq)
- URI and connstring forms already worked; CLI flags now work too
- `-p` flag accepts comma-separated port list (single port still works — fully backward-compatible)
- `resolve_hosts()` extended with priority 3: CLI multi-host (after URI and conninfo)

## Tests

5 new unit tests in `src/connection.rs`:
- `test_resolve_params_cli_multihost_single_port` — single port applied to all hosts
- `test_resolve_params_cli_multihost_per_host_ports` — per-host port assignment
- `test_resolve_params_cli_multihost_port_reuse` — last port reused when fewer ports than hosts
- `test_resolve_params_cli_multihost_positional` — positional HOST arg with commas
- `test_resolve_params_cli_single_host_regression` — single-host path unchanged

3 new integration tests in `tests/connection_paths.rs` (Group I, run via connection-tests CI):
- `i1_multihost_first_fails_second_succeeds` — failover to second host
- `i2_multihost_all_fail` — error message when all hosts unreachable
- `i3_single_host_regression` — single host still connects normally

Existing `f7_multihost_failover` canary test (issue #724) now exercises the same code path.

All 1814 unit tests pass.


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/rpg#751 - fix/743-multihost: multi-host failover via CLI flags
- **Branch:** `fix/743-multihost`
- **Commits:** 3 (feat + 2 fix/lint/docs)

---

### BLOCKING ISSUES (1)

**MEDIUM** `src/connection.rs:893-897` - Early return on invalid port leaves `params.hosts` stale
> When `-p` contains an invalid port token (e.g. `-h h1,h2 -p 5432,notaport`), `port_parse_ok` is set to `false` and the function returns early with `return`. The comment says "Fall through to the single-host path", but `return` exits `resolve_hosts` entirely — it does **not** reach the fallback `params.hosts = vec![(params.host.clone(), params.port)]` at line 919.
>
> At the point of the early return, `params.hosts` still holds the value from `ConnParams::default()` (i.e., the OS socket-based default host/port from `default_host_port()`), not the resolved `params.host`/`params.port`. So the caller ends up with a mismatch: `params.host` = resolved host, `params.port` = resolved port, but `params.hosts` = stale default. Depending on which field the connection layer actually iterates, this can silently connect to the wrong host.
>
> The test `test_resolve_params_cli_multihost_invalid_port_does_not_silently_use_wrong_port` only asserts `port != 0`, which is satisfied by the stale default (e.g. 5432), not detecting the mismatch.
>
> **Fix:** Replace `return;` with the fallback assignment before returning:
> ```rust
> if !port_parse_ok {
>     // Fall through to single-host path.
>     params.hosts = vec![(params.host.clone(), params.port)];
>     return;
> }
> ```
> Also strengthen the invalid-port test to assert that `params.hosts[0].0 == "h1"` and `params.hosts[0].1 == 5432` (the resolved values), not just that port ≠ 0.

---

### NON-BLOCKING (2)

**LOW** `src/connection.rs:872-874` - `port_str` vs `port_pos` precedence may shadow per-host ports in positional form

> For positional args (no `-h`/`-p` flags), `port_pos` carries the raw port string. The code at line 872 picks `port_str` first, then falls back to `port_pos`. However, `port_str` is populated from the named `-p` flag (see `main.rs:329`), while `port_pos` holds the positional PORT argument. If a user types `rpg h1,h2 7777,7778` (positional form), `o.port_str` will be `None` and `o.port_pos` will be `Some("7777,7778")`, which is correct. The test `test_resolve_params_cli_multihost_positional` confirms this path works. No active bug, but the fallback chain is non-obvious. A clarifying comment at the fallback point would help future maintainers.

**LOW** `src/main.rs:320-322` - Silent parse failure for first port in `-p`

> `conn_opts()` extracts `port_u16` by splitting on `,` and parsing the first token:
> ```rust
> let port_u16: Option<u16> = self.port.as_deref()
>     .and_then(|s| s.split(',').next())
>     .and_then(|s| s.trim().parse().ok());
> ```
> If the first token is not a valid u16 (e.g. `-p notaport,5432`), `port_u16` becomes `None` silently. The error surfaces later as a confusing "could not connect" rather than "invalid port". Consider emitting a user-facing warning or validation error at this layer, consistent with the explicit `eprintln!` guard in `resolve_hosts`.

---

### POTENTIAL ISSUES (2)

**MEDIUM** `src/connection.rs:860-861` - Named `-h` and positional HOST are treated as equivalent for multi-host; priority vs conninfo is not tested *(confidence: 6/10)*

> The comment at line 808-811 documents that CLI `-h` takes priority over single-host conninfo. However, there is no unit test covering the scenario where both `opts.host` (CLI `-h host1,host2`) and a single-host conninfo are present simultaneously. If a user has `PGHOST=dbserver` in a `.pg_service.conf` and passes `-h h1,h2`, the intent is clearly CLI wins — but without a test this priority path is unverified.

**LOW** `tests/connection_paths.rs:1513-1625` - Integration tests i1/i2 use a hard-coded non-routable hostname, which may resolve differently under different DNS configs *(confidence: 5/10)*

> `bad-host-xyz-nonexistent` is used as the unreachable first host. On some corporate or CI environments with wildcard DNS, this may resolve to a real IP. The test would then fail for a different reason (connection refused vs. name resolution failure). Using `192.0.2.1` (TEST-NET, RFC 5737) as the unreachable address is more reliable for tests that need a guaranteed-unreachable host.

---

### Summary

| Security | 0 | 0 | 0 |
| Bugs | 1 | 1 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 1 | 0 | 0 |

Notes:
- **Findings**: High-confidence issues (8-10/10) — blocking or non-blocking per severity
- **Potential**: Medium-confidence issues (4-7/10) — review manually
- **Filtered**: Low-confidence issues (0-3/10) — excluded as likely false positives

---

*REV-assisted review (AI analysis by [postgres-ai/rev](https://gitlab.com/postgres-ai/rev))*

NikolayS:
## REV Code Review — PR #751

**PR:** NikolayS/rpg#751 — feat(connection): multi-host failover
**Branch:** fix/743-multihost
**AI-Assisted:** Yes

---

### BLOCKING ISSUES

**HIGH** `src/connection.rs:870` — `params.host.clone_from(&host_list[0].0)` — unchecked index on `host_list` *(confidence: 9/10)*
> `host_list` is populated only when `host_parts.len() > 1`, so it has at least 2 elements at the point `host_list[0]` is accessed. This is safe — the guard is `host_parts.len() > 1` at line 858. However, this implicit invariant is fragile: if the guard is ever relaxed (e.g., `>= 1`), `host_list[0]` will panic on empty input. Replace with a `host_list.first()` pattern and an explicit `if host_list.is_empty() { return; }` guard to make the invariant explicit and panic-safe.

> **Fix:**
> ```rust
> if let Some(first) = host_list.first() {
>     params.host.clone_from(&first.0);
>     params.port = first.1;
> }
> params.hosts = host_list;
> ```

---

### NON-BLOCKING

**LOW** `src/main.rs:321` — `port_u16` parses only the first token of `-p` for backward-compat callers
> The comment explains this correctly. The only concern is that if `-p 5432,notaport` is supplied, `port_u16` is `Some(5432)` while `port_str` is `"5432,notaport"` — the invalid-port detection in `resolve_hosts` handles the mismatch. This is correct behavior.

**LOW** `src/connection.rs:854–858` — `cli_host` prefers `o.host` over `o.host_pos` without a comment explaining priority
> Document: `o.host` is the named `--host` flag; `o.host_pos` is the positional HOST argument. Named flag takes priority, consistent with single-host resolution elsewhere.

**LOW** `src/connection.rs:804–806` — doc comment for `resolve_hosts` priority list is now 4 items; previously 3
> The updated comment correctly documents all 4 priority tiers. Minor: item 3 says "CLI -h flag (takes priority over single-host conninfo)" — this is only true when `-h` supplies *multiple* hosts. Clarify: "Multi-host from CLI -h flag when multiple hosts specified".

---

### POTENTIAL ISSUES

**MEDIUM** `src/connection.rs:880–883` — port reuse uses `last_port` (last seen port) for subsequent hosts
> When fewer ports than hosts are provided (e.g., `-p 5432,5433` with 3 hosts), host 3 gets port 5433 (last seen). This matches PostgreSQL libpq behavior and the test `test_resolve_params_cli_multihost_port_reuse` validates it. Confirm this is intentional and matches libpq semantics — it does. No action needed beyond the existing test.

**MEDIUM** `tests/connection_paths.rs` — I1/I2/I3 integration tests are all `#[ignore]` *(confidence: 7/10)*
> The tests are correctly ignored for standard CI (they require a live Postgres). Confirm there is a `connection-tests` CI job that runs them with `cargo test -- --ignored`; if not, these tests will never run automatically and are effectively dead code.

---

### Summary

| Security | 0 | 0 | 0 |
| Bugs | 1 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 2 | 0 |
| Docs | 1 | 0 | 0 |

The core logic is solid — URI and conninfo multi-host paths are unchanged; the new CLI path is additive and well-guarded. The `port_str` design cleanly preserves backward compatibility. Unit test coverage is comprehensive (6 new tests). The BLOCKING finding is a low-risk latent panic: the current guard makes it unreachable, but it should be made explicit.

CI is currently red (cargo fmt only — fixed in latest push). Re-run expected to pass.

---

*REV-assisted review (AI analysis by [postgres-ai/rev](https://gitlab.com/postgres-ai/rev))*

NikolayS:
REV

## #751 — feat(connection): multi-host failover

**This is the most consequential PR of the batch.** Connection handling is a correctness-critical path. The implementation is generally sound but has a few things worth checking carefully.

### What's good

- `port_str: Option<String>` as a raw field alongside the parsed `port: Option<u16>` is the right way to handle comma-separated port lists at the CLI layer — clean separation of concerns
- Port reuse for remaining hosts (last port carries forward) matches libpq behavior
- Invalid port early-return path is defensive and correct: fail loudly rather than silently misassign ports
- Test coverage is thorough: single port, per-host ports, port reuse, positional HOST argument, invalid port

### Concerns

1. **CLI multi-host priority vs conninfo** — the doc comment says "CLI -h takes priority over single-host conninfo" which is the intended behavior. But what happens when `-h h1,h2` is given AND the conninfo string also has multi-host (`host=h1,h2`)? The conninfo path runs first and populates `params.hosts`, then the CLI path overwrites it. Is that the right precedence? A test for this case would confirm the behavior is intentional.

2. **`eprintln!` for invalid port** — printing to stderr from deep in `resolve_hosts` is unexpected for a library-style function. This should either return an error (propagate up) or use the logging subsystem. The current approach makes it hard to test and breaks the no-side-effects contract of a resolver function.

3. **`port_parse_ok = false` but `port_parts` is still partially populated** — when a port is invalid, `port_parse_ok` is set false and the function falls back to single-host. But the loop continues collecting (with 0 as the invalid value). This is fine since the result is discarded, but the code is confusing to read. A cleaner approach: early-return from the closure on first failure.

4. **IPv6 host support** — `h.split(',').map(str::trim)` will split on commas inside IPv6 bracket notation (`[::1]`). Multi-host IPv6 via CLI is probably rare but worth a note in the code or a test asserting the behavior ("IPv6 hosts must use URI format for multi-host").

5. **`params.hosts = vec![(params.host.clone(), params.port)]`** — the fallback on invalid port sets hosts to the already-resolved single host. But if the user supplied `-h h1,h2` and an invalid port, they clearly intended multi-host. Silently falling back to single-host could cause a confusing connection to just h1 when the intent was failover. At minimum the error message should make clear that only h1 will be used.

### Minor

- Doc update in `connection.rs` header: "Multiple hosts may be specified via:" with CLI flags documented — good addition.
- The `resolve_hosts` function signature now takes `opts: Option<&CliConnOpts>` — Option wrapping is correct since this is also called from tests that don't have CLI opts.

**Verdict: approve with caveats. Issue 2 (eprintln in resolver) is worth cleaning up. Issue 4 (IPv6) should be documented. The rest are acceptable for now.**

NikolayS:
CI is green and REV review found no blocking findings. This PR is ready for merge — awaiting Nik's approval.

PR #755, opened 2026-03-26
feat(repl): establish /cmd convention for rpg-specific commands

### LINKED ISSUE #754: feat: establish /cmd convention for all rpg-specific commands
## Summary

Establish a clear command namespace convention:

- `\` = psql-compatible commands only (anything psql has: `\d`, `\l`, `\dt`, `\copy`, `\s`, etc.)
- `/` = everything rpg adds on top (AI, monitoring, diagnostics, all rpg-specific extensions)

## Motivation

The current split is ambiguous — some rpg-specific commands use `\` which blurs the boundary. The new convention makes it immediately clear to users:

> "If psql has it, it's `\`. If it's rpg-specific, it's `/`."

This also keeps `alias psql=rpg` working perfectly since psql users only type `\` commands.

### Command renames
Audit all rpg-specific commands and move any that aren't in psql to the `/` namespace. Known cases:
- Any diagnostics commands (`\dba` etc.) — check if psql has them; if not, rename to `/`
- `\ash` (issue #753) → `/ash` (already rpg-specific, not in psql)

### Code changes
- Update command dispatch in `src/repl/mod.rs` for any renamed commands
- Update help text (`\?` output and any `/help` equivalent)
- Update CHANGELOG
- Update README if commands are documented there

### Documentation & assets
- Add a note in CLAUDE.md or a COMMANDS.md explaining the convention so it's enforced going forward
- Update all examples in README.md that reference renamed commands
- Re-record any GIFs/screenshots that show old command names (recording is automated via tmux pipeline)
- Update any inline code comments referencing renamed commands

## Acceptance criteria
- [ ] All rpg-specific commands use `/` prefix
- [ ] All psql-compatible commands keep `\` prefix
- [ ] Help text reflects the new convention
- [ ] `alias psql=rpg` still works (no `\` commands broken)
- [ ] Convention documented for contributors (CLAUDE.md or COMMANDS.md)
- [ ] README examples updated
- [ ] GIFs / screenshots re-recorded via tmux pipeline

NikolayS:
Closes #754. Moves rpg-specific commands to / namespace, keeps psql-compatible commands as \. See issue for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Single binary, no dependencies, cross-platform.
 
 - **psql-compatible** — drop-in replacement (`\d`, `\dt`, `\copy`, `\watch`, ...)
 - **AI assistant** — `/ask`, `/fix`, `/explain`, `/optimize`
-- **DBA diagnostics** — 15+ `\dba` commands for activity, locks, bloat, indexes
+- **DBA diagnostics** — 15+ `/dba` commands for activity, locks, bloat, indexes
 - **Schema-aware completion** — tab completion for tables, columns, keywords
 - **TUI pager** — scrollable pager for large result sets
 - **Syntax highlighting** — SQL keywords, strings, operators; color-coded errors (red), warnings (yellow), notices (cyan)
@@ -79,12 +79,12 @@ select * from orders where status = 'pending';
 | `/compact [focus]` | Compact conversation context (optional focus topic) |
 | `/budget` | Show token usage and remaining budget |
 
-### \text2sql — natural language to SQL
+### /text2sql — natural language to SQL
 
 By default, the generated SQL is shown in a preview box and you confirm before it runs:
 
 ```
-postgres=# \text2sql
+postgres=# /text2sql
 Input mode: text2sql
 postgres=# what is DB size?
 ┌── sql
@@ -97,14 +97,14 @@ Execute? [Y/n/e]
 (1 row)
 ```
 
-### \yolo — fast natural-language mode
+### /yolo — fast natural-language mode
 
-`\yolo` combines text2sql and silent auto-execute in one command: it enables
+`/yolo` combines text2sql and silent auto-execute in one command: it enables
 text2sql input, hides the SQL preview box, and executes immediately without
 confirmation.
 
 ```
-postgres=# \yolo
+postgres=# /yolo
 Execution mode: yolo
 postgres=# what is DB size?
  db_size
@@ -113,12 +113,12 @@ postgres=# what is DB size?
 (1 row)
 ```
 
-Toggle back with `\sql` or `\interactive`. Show/hide the SQL preview box with
+Toggle back with `/sql` or `/interactive`. Show/hide the SQL preview box with
 `\set TEXT2SQL_SHOW_SQL on`.
 
-![\\t2s text-to-SQL mode and \\yolo auto-execute mode in action](demos/gif3_t2s.gif)
+![/t2s text-to-SQL mode and /yolo auto-execute mode in action](demos/gif3_t2s.gif)
 
-*`\t2s` shows a SQL preview with confirmation; `\yolo` skips the preview and executes immediately.*
+*`/t2s` shows a SQL preview with confirmation; `/yolo` skips the preview and executes immediately.*
 
 ### /fix — auto-correct errors
 
@@ -169,9 +169,9 @@ postgres=# /optimize
 Upload the last EXPLAIN plan to an external visualizer:
 
 ```
-\explain share depesz     → posts to explain.depesz.com
-\explain share dalibo     → posts to explain.dalibo.com
-\explain share pgmustard  → posts to pgMustard (requires PGMUSTARD_API_KEY)
+/explain-share depesz     → posts to explain.depesz.com
+/explain-share dalibo     → posts to explain.dalibo.com
+/explain-share pgmustard  → posts to pgMustard (requires PGMUSTARD_API_KEY)
 ```
 
 ### EXPLAIN display format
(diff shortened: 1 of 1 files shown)


NikolayS:
# REV Code Review — PR #755: feat(repl): establish /cmd convention for rpg-specific commands

**CI:** FAILING (rustfmt) | **Diff scope:** `CLAUDE.md`, `COMMANDS.md` (new), `src/repl/ai_commands.rs`

---

## Blocking findings

**[CI / BLOCKING] rustfmt failure on `src/repl/ai_commands.rs`**

The lint step fails at `cargo fmt --check` with:

```
-            crate::dba::execute(client, &format!("ash {rest}").trim().to_owned(), false, Some(&caps), settings).await;
+            &format!("ash {rest}").trim().to_owned(),
```

The `/ash` handler has a long `crate::dba::execute(...)` call that rustfmt wants broken across multiple lines. Run `cargo fmt` on the branch to fix. This is the only blocker.

---

## Non-blocking findings

**[BUG / MEDIUM] `dispatch_ai_command` returns `Some(MetaResult)` for mode-change commands but the handle_line caller only handles `Reconnected` and `Quit`**

In `handle_line`:

```rust
if let Some(result) = dispatch_ai_command(...).await {
    match result {
        MetaResult::Reconnected(c, p) => return HandleLineResult::Reconnected(c, p),
        MetaResult::Quit => return HandleLineResult::Quit,
        _ => {}
    }
}
```

The `/sql`, `/text2sql`, `/plan`, `/yolo`, `/interactive`, `/mode` branches inside `dispatch_ai_command` call `apply_mode_change` or print to stderr and return `None`. They do **not** return a `MetaResult` — they handle the state change internally and return `None`. This is correct. However, the session-related branches (`/session resume`) return `Some(MetaResult::Reconnected(...))` via `dispatch_session_resume`. Confirm that all paths that need to propagate a result do so via `Some(...)` and that `apply_mode_change` side effects in settings are sufficient (they are, since `settings` is `&mut`). This appears correct but the control flow is subtle enough to warrant a comment.

**[BEHAVIOR / LOW] `/refresh` inside `dispatch_ai_command` prints `\refresh is deprecated` on the `\refresh` path but `/refresh` in the new `/` path would hit the `MetaCmd::RefreshSchema` arm, not the AI dispatch**

The deprecation eprintln on `MetaCmd::RefreshSchema` is in `dispatch_meta` (the `\` path). The new `/refresh` handler in `dispatch_ai_command` does not have a deprecation message — which is correct. But check that `/refresh` is actually routed to `dispatch_ai_command` and not accidentally intercepted by the metacmd parser (which handles `\` commands). If the REPL parser matches any `/`-prefixed token as a metacmd, `/refresh` could double-execute.

**[STYLE / LOW] `is_budget_exempt` list in `dispatch_ai_command` will need updating for every new `/` command**

The long `||` chain is already 20+ conditions. Consider inverting: define a `requires_ai_budget` function that returns `true` only for `/ask`, `/fix`, `/explain`, `/optimize`, `/describe`, and gate budget checking there. Everything else is exempt by default.

**[DOCS / LOW] COMMANDS.md lists `/n <name> [args...]` but the `/n+`, `/nd`, `/np` commands use `+`, `d`, `p` suffixes — should be clear these are distinct commands**

The table reads naturally but a new user might not realize `/n+` is a standalone command, not `/n` with argument `+`. A footnote or note in the rationale section would help.

---

## Potential issues (low confidence)

**[BEHAVIOR] Deprecation messages go to stderr via `eprintln!`**

All `\cmd is deprecated; use /cmd instead.` messages use `eprintln!`. This is correct for interactive use. However, if a user pipes rpg output or uses `-c`, the deprecation warning on stderr may intermix with query output in unexpected ways. This matches how psql handles warnings, so it is acceptable.

---

## Summary

Good architectural change — the `\` vs `/` namespace split is clear and the `COMMANDS.md` reference doc is useful. The only blocker is the rustfmt failure on the `/ash` handler call. Fix agent should run `cargo fmt` and push. Once CI is green, the PR is solid.

NikolayS:
## REV Re-Review — PR #755 (after fixes)

**Branch:** `feat/754-slash-convention`
**Head:** `7312985`
**Commits since REQUEST_CHANGES:** `5694b29`, `7312985`

---

### BLOCKING ISSUES

None. Both blocking bugs from the prior review are resolved.

---

### Fix verification

**B1 — `/explain-share` dead code (RESOLVED)**

`/explain-share` handler is now at line 429, before the `/explain` handler at line 436. The `starts_with("/explain")` match no longer swallows `/explain-share depesz`. Fix is correct.

**B2 — `/n*` prefix collision (RESOLVED)**

All four dispatch branches now use the `input == "/ns" || input.starts_with("/ns ")` pattern:
- `/ns` — line 647
- `/nd` — line 688
- `/np` — line 709
- `/n` — line 722

`/ndrop` will no longer silently route to `/nd`. Fix is correct.

**README update (requested as NON-BLOCKING — done)**

All `\dba`, `\text2sql`, `\yolo`, `\sql`, `\interactive`, `\commands`, `\explain share` references updated to `/dba`, `/text2sql`, `/yolo`, `/sql`, `/interactive`, `/commands`, `/explain-share`. The 25+ stale references are addressed.

---

### NON-BLOCKING observations (new diff only)

**LOW — `/explain-share` bare call prints usage, not the more helpful "no plan to share" message**

When called as `/explain-share` (no service arg), the handler prints `Usage: /explain-share <depesz|dalibo|pgmustard>`. That's fine. But the prior behavior also handled the empty-service case via `dispatch_explain_share`, which would check `settings.last_explain_text` first and print a more contextual error. Current behavior is acceptable — minor UX detail, not blocking.

**INFO — `/explain-share` still accepts partial prefix match**

`/explain-shareXYZ` (no space, garbage suffix) would still match `strip_prefix("/explain-share")` and pass `"XYZ"` as the service. The service matching inside `dispatch_explain_share` will reject it with an unknown-service error. Not a bug, just noting the behavior is consistent with the rest of the dispatch chain.

---

### Summary

| Commit | Change | Status |
|---|---|---|
| `5694b29` | Move `/explain-share` before `/explain`; fix `/n*` prefix collisions | Correct |
| `7312985` | Update README to use `/` prefix throughout | Complete |

CI is green. No blocking findings. **PR #755 is ready for merge — awaiting Nik's approval.**

PR #762, opened 2026-03-27
feat(ash): pg_ash history integration — pre-populate /ash timeline from ash.samples (#761)

### LINKED ISSUE #761: feat(ash): pg_ash history integration — /ash reads ash.samples when pg_ash is installed
## Goal

When `pg_ash` is installed on the server, `/ash` should immediately show the full history available in `ash.samples`, then continue streaming live data. Today it always uses the live ring buffer (600 samples max, ~10 min).

## Current behavior

`AshState.pg_ash_installed` is detected but `ViewMode::History` never activates — the code falls back to live ring buffer unconditionally (documented TODO in `state.rs:102`).

## Desired behavior

1. **On connect** — detect `pg_ash` via `SELECT 1 FROM pg_extension WHERE extname = 'pg_ash'`
2. **If installed** — query `ash.samples` for the configured window (default 10 min, up to 24h via zoom)
3. **Merge** — history snapshots pre-populate the ring buffer; live `pg_stat_activity` polls append new samples
4. **Seamless** — left side of timeline = history, right side = live. No gap, no jump.
5. **Zoom** — when user zooms out beyond ring buffer capacity, re-query `ash.samples` for larger window

### Sampler changes (`src/ash/sampler.rs`)

Add `query_ash_history()`:
```sql
select
    wait_event_type,
    wait_event,
    query_id,
    left(query, 80) as q,
    count(*) as count,
    bucket_ts  -- truncated to bucket_secs
from ash.samples
where sample_ts >= now() - interval '10 minutes'
group by 1, 2, 3, 4, bucket_ts
order by bucket_ts
```

Note: `ash.samples` schema — verify column names against pg_ash v1.2 (`sample_ts`, `wait_event_type`, `wait_event`, `query_id`, `query`).

### State changes (`src/ash/state.rs`)

Activate `ViewMode::History` when `pg_ash_installed = true`. On zoom out past ring buffer, trigger re-query.

### mod.rs event loop

```rust
// On startup, if pg_ash installed:
let history = query_ash_history(&client, window_secs).await?;
for snap in history {
    snapshots.push_back(snap);
}
// Then continue live polling as normal
```

### CI tests required

- Unit test: `query_ash_history()` SQL correctness (mock rows)
- Integration test (ignored by default, needs pg_ash): populate `ash.samples`, verify ring buffer pre-populated
- State test: `pg_ash_installed = true` activates history mode

### AI test required

New file: `tests/ai/slash-ash-pgash.md`
- Setup: install pg_ash, run pgbench for 5 min to populate history
- Steps: connect, run `/ash`, verify left side of timeline shows pre-existing history (not empty)
- GIF: record 10s showing bars from minute 0, new bars scrolling in live
- Post GIF to PR as evidence

## References

- pg_ash repo: https://github.com/NikolayS/pg_ash
- ash.samples schema: see pg_ash v1.2 `sql/` directory
- Current stub: `src/ash/state.rs:102` (`TODO: history mode (pg_ash Layer 2)`)
- PR #756 (base implementation): https://github.com/NikolayS/rpg/pull/756

## Scope

New PR on top of main (after #756 merges). Do not modify existing /ash live behavior — history is additive.

NikolayS:
Closes #761.

## What this does

When `pg_ash` is installed on the server, `/ash` now pre-populates the timeline with historical data from `ash.wait_timeline()` before starting the live polling loop. Falls back to live-only silently when pg_ash is absent.

## How it works

On `/ash` startup:
1. `detect_pg_ash()` checks if pg_ash is installed (existing behavior)
2. If installed: call `query_ash_history(client, 600)` — queries `ash.wait_timeline('600 seconds'::interval, '1 second'::interval)`
3. Results (oldest-first `AshSnapshot`s) pre-fill the ring buffer
4. Live `pg_stat_activity` polling continues as normal — new samples scroll in on the right

The left side of the timeline shows history immediately; the right side grows with live data.

## Key design decisions

- **Graceful degradation**: `query_ash_history()` returns empty vec on any error — no crash, no error spam
- **`ash.wait_timeline()`**: uses the public pg_ash API rather than decoding the internal `int[]` encoding directly
- **Wait event parsing**: `wait_timeline` returns `"Type:Event"` strings; parsed into `(wtype, wevent)` pairs matching live sampler format
- **Ring buffer cap**: history respects the same 600-sample cap as live mode

## Tests

1905 tests passing (26 new: history parsing, ring buffer capacity, graceful degradation, wait event format parsing)

## Still needed (AI test + GIF)

Per our convention, a `tests/ai/slash-ash-pgash.md` file with recording instructions will be added once pg_ash is available in the test environment.


NikolayS:
## REV findings addressed — commit `b822901`

All blocking findings from the REV fixed:

**M1 (FIXED):** Replaced `format!()` SQL interpolation with parameterized `$1` in `query_ash_history_inner`. Renamed helper from `_interval` → `_inner` (no longer keyed on string interval). Zero injection surface now.

**M2 (FIXED):** `samples::int8` (was `::int4`), row read as `i64`, saturating to `u32::MAX` instead of silently wrapping.

Non-blocking addressed:
- **L3:** Doc comment on `pg_ash_installed` cleaned up (double-sentence removed)
- **L2:** `retention_seconds` doc clarified as reserved/not yet wired
- **L4:** Comment added explaining CPU ref line populates on first live snapshot
- **M3:** TODO comment added for history-mode per-frame caching

Remaining deferred (L1 - `pg_ash_installed` dead code): field retained for future status-bar indicator; `#[allow(dead_code)]` stays. W1 test duplication noted in REV — acceptable until logic stabilizes.

**Test results:** 1864 passed, 0 failed ✓  
**Clippy:** clean ✓  
**fmt:** clean ✓

PR #770, opened 2026-03-29
feat(ash): apply statement_timeout to live pg_stat_activity queries

### LINKED ISSUE #769: feat(ash): apply statement_timeout to live pg_stat_activity queries
## Problem

`/ash` live mode queries `pg_stat_activity` on every tick with no `statement_timeout`. On clusters with thousands of connections the query can take hundreds of milliseconds, causing TUI lag and sample interval drift. If it ever hangs, the entire TUI blocks.

pg_ash solves the same problem by setting `statement_timeout = '500ms'` on its pg_cron job — fail fast rather than pile up.

## Proposed fix

In `src/ash/sampler.rs`, wrap each `live_snapshot()` call with:

```sql
SET LOCAL statement_timeout = '500ms';
```

before the `pg_stat_activity` query. If it times out:
- Record the tick as a missed sample (increment a counter)
- Display a brief indicator in the TUI (e.g. `[missed]` on the status bar)
- Continue — do not crash or block

## Reference

pg_ash uses exactly this pattern: `set statement_timeout = '500ms'; select ash.take_sample()` as the pg_cron command. See [ash-install.sql line 796](https://github.com/NikolayS/pg_ash/blob/main/sql/ash-install.sql#L796).

NikolayS:
## Problem

`/ash` live mode queries `pg_stat_activity` on every tick with no `statement_timeout`. On clusters with thousands of connections this can take hundreds of milliseconds, causing TUI lag and sample interval drift.

## Fix

Sets `statement_timeout = '500ms'` before each live `pg_stat_activity` query — the same guard pg_ash uses on its per-second cron job.

If the query times out:
- Tick is skipped (`LiveSnapshotResult::Missed`)
- `missed_samples` counter incremented in `AshState`
- Status bar shows `missed: N` when non-zero

## Changes

- `src/ash/sampler.rs`: new `LiveSnapshotResult` enum; `SET LOCAL statement_timeout` before query; return `Missed` on `QUERY_CANCELED`
- `src/ash/mod.rs`: update caller to handle `Ok/Missed/Err` variants
- `src/ash/state.rs`: add `missed_samples: u32` field
- `src/ash/renderer.rs`: show `missed: N` in status bar when non-zero

Closes #769


NikolayS:
## REV Code Review — PR #770 (supplemental, commit `bbdc017`)

**PR:** NikolayS/rpg#770 — feat(ash): apply statement_timeout to live pg_stat_activity queries
**Scope:** commit `bbdc017` only — fix(ash): drain key events within interval
**File changed:** `src/ash/mod.rs`

> Previous review (`0c1f073`) stands — no new issues found there. This covers the new commit only.

---

### BLOCKING ISSUES (0)

None.

---

### NON-BLOCKING (2)

**LOW** `src/ash/mod.rs` — inner event-drain loop uses `saturating_duration_since` but never re-checks if deadline already passed before the first `event::poll` call

If `live_snapshot` takes longer than `refresh_interval_secs` (e.g., slow query, 500ms timeout approaching the 1s interval), `deadline` may already be in the past when the event loop begins. `remaining.is_zero()` catches this correctly on the first iteration — the loop exits immediately and no events are drained. This is the right behavior, but a one-line comment would clarify the intent for reviewers who might wonder why there's no pre-check before the loop.

**LOW** `src/ash/mod.rs` — `terminal.draw()` called up to 3 times per tick in worst case

Per tick: (1) initial draw after snapshot, (2) draw after Enter/drill-in, (3) draw after handle_key. In a tight key-hold scenario the inner loop can call `draw` many times per second. In practice this is fine — ratatui's double-buffer diff means no-op redraws are cheap. Worth noting in a comment for future contributors who might wonder why the draw count isn't bounded.

---

### POTENTIAL ISSUES (0)

None.

---

### Summary

| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 2 | 0 | 0 |

The event-drain design is correct and idiomatic for TUI event loops. The `'outer` label is clean, wall-clock anchoring is solid, and the immediate-redraw pattern for drill-in and selection changes is the right UX choice.

**Overall:** No blockers. Combined with the prior APPROVE on `0c1f073`, PR #770 remains ready for merge.

---

*REV-assisted review · [postgres-ai/rev](https://gitlab.com/postgres-ai/rev)*

PR #784, opened 2026-04-04
fix: improve PostgreSQL regression test compatibility

NikolayS:
## Summary

Fixes a large number of failures in PostgreSQL's regression test suite run against rpg, increasing PASS rate from ~120 to ~148 out of 245 tests.

### Key fixes

**`\d` / describe output:**
- Add proper `\d` output for indexes: `Key?`, `Definition`, `Storage`, `Stats target` columns with correct footer line (`btree, for table "..."`)
- Reorder partition sections: `Partition key` / `Partition of` / `Partition constraint` now appear **before** the `Indexes:` section (matching psql)
- Show `Partition constraint:` only in `\d+` mode
- Show `Number of partitions: N (Use \d+ to list them.)` for `\d` non-plus; show full list only for `\d+`
- Fix `Partitions:` list ordering: order by `c2.oid::regclass::text` (alphabetical by name, so `part_null` < `part_p1`)
- Use `pg_table_is_visible()` for unqualified partition names in `Partitions:` list
- Use `pg_get_partition_constraintdef()` instead of querying `pg_constraint` directly
- Fix index column extraction: use `find('(')` not `rfind('(')` to correctly capture expression indexes like `((b = 1))`
- Show `View definition:` only for `\d+` (not plain `\d`)
- Remove extra space prefix from view definition lines (pg_get_viewdef already provides correct indentation)
- Fix `row.get(1)` → `row.get(0)` in inherits query (was causing a panic)

**Error output:**
- Add `CONTEXT:` field from `DbError::where_()` to error messages
- Fix error field ordering: Original position → DETAIL → HINT → CONTEXT → QUERY/position

**Echo/REPL:**
- Echo original (pre-interpolation) line so `:varname` references appear unexpanded
- Echo blank lines only inside string literals (not after comment-only buffers)
- Add `\;` multi-command separator: splits SQL on same line, stops chain on first error
- Mirror `\set ECHO` to `settings.echo_all`; `\set QUIET` to `settings.quiet`
- Add DML RETURNING command tag output (INSERT/UPDATE/DELETE N after result set)

**`\gexec` in script mode:**
- `GExecBuffer` was missing from `exec_lines()` match, causing `\gexec` to be silently ignored in non-interactive mode. Buffer would accumulate leading to syntax errors on subsequent queries.
- Suppress `\gexec` command tag output when `ECHO=none`

**`\df` variant filtering:**
- Added `kind_filter: Option<char>` to `ParsedMeta` for `\dfn`/`\dfp`/`\dfa`/`\dfw` filtering

## Test plan

- [ ] Run `tests/compat/test-psql-regress.sh` — 148+ tests now passing (was ~120)
- [ ] Verify triggers, cluster, create_table, create_view tests pass
- [ ] Confirm no regressions on previously passing tests

https://claude.ai/code/session_013GRAngqk75khcjewoEwzm3


NikolayS:
Superseded by #788 which combines this with #787 into a single reviewed PR.

PR #801, opened 2026-04-12
fix(compat): make test-psql-regress.sh portable to bash 3.2 (#797)

### LINKED ISSUE #797: test-psql-regress.sh: bash 3.2 portability (macOS)
## Problem
`tests/compat/test-psql-regress.sh` uses `declare -A` (associative arrays) which requires bash 4+. macOS ships bash 3.2 by default, causing the script to fail with a syntax error.

## Options
1. Replace `declare -A` with a bash 3.2 compatible alternative (parallel arrays or grep-based lookup)
2. Add a version check at the top with a clear error message
3. Change shebang to explicitly require bash 5 (e.g., `/opt/homebrew/bin/bash`)

## Acceptance criteria
- Script either works on bash 3.2 or fails with a clear error message
- Document the bash 5 requirement if not made portable

NikolayS:
## Summary

- Replace `declare -A SEEN_TESTS=()` (associative arrays, bash 4+ only) with a portable pipe-delimited string and `case`-based lookup functions (`_is_seen` / `_mark_seen`) that work on bash 3.2
- Add a non-fatal warning near the top of the script when bash < 4 is detected

## What was changed

`tests/compat/test-psql-regress.sh` used `declare -A` at line 477 to create an associative array for test name deduplication. macOS ships bash 3.2 by default, which does not support associative arrays, causing a syntax error.

The fix replaces the associative array with a simple string variable `_seen_tests` that stores seen names in a `|name1|name2|...` format. Two helper functions provide the same dedup interface:
- `_is_seen NAME` — returns 0 if the name has been recorded, 1 otherwise (uses `case` glob matching)
- `_mark_seen NAME` — appends the name to the tracking string

This approach is O(n) per lookup in the worst case but entirely sufficient for the number of regression tests involved, and is fully portable to bash 3.2.

## Test plan

- [x] `bash -n tests/compat/test-psql-regress.sh` passes (no syntax errors)
- [ ] CI passes (GitHub Actions)
- [ ] Verify dedup still works: tests from `parallel_schedule` are not duplicated when also found as `.sql` files

Closes #797

### DIFF
--- a/tests/compat/test-psql-regress.sh
+++ b/tests/compat/test-psql-regress.sh
@@ -3,6 +3,13 @@
 set -Eeuo pipefail
 IFS=$'\n\t'
 
+# Warn (but don't fail) if running on bash < 4. The script is portable to
+# bash 3.2, but newer bash is recommended for better performance.
+if [[ "${BASH_VERSINFO[0]}" -lt 4 ]]; then
+  echo "WARNING: bash ${BASH_VERSION} detected." \
+    "bash 4+ is recommended but not required." >&2
+fi
+
 # On macOS, GNU coreutils installs as 'gtimeout'; add gnubin to PATH if present.
 for _gnubin in \
   /opt/homebrew/opt/coreutils/libexec/gnubin \
@@ -474,7 +481,21 @@ setup_regress_db
 
 # Collect the ordered list of tests from parallel_schedule.
 declare -a ORDERED_TESTS=()
-declare -A SEEN_TESTS=()
+
+# Portable deduplication for bash 3.2+ (no associative arrays).
+# _seen_tests holds "|name1|name2|…|" so we can match with case.
+_seen_tests="|"
+
+_is_seen() {
+  case "${_seen_tests}" in
+    *"|${1}|"*) return 0 ;;
+    *)          return 1 ;;
+  esac
+}
+
+_mark_seen() {
+  _seen_tests="${_seen_tests}${1}|"
+}
 
 schedule_file="${PG_SRC}/src/test/regress/parallel_schedule"
 if [[ -f "${schedule_file}" ]]; then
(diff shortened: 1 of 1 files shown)


NikolayS:
## REV Review — PR #801

**Verdict:** PASS

### Findings

- **[POTENTIAL]** `case` glob matching in `_is_seen` — The function uses `case "${_seen_tests}" in *"|${1}|"*)` which treats `$1` as a glob pattern. If a test name ever contained `*`, `?`, or `[...]`, it could false-match other names. In practice, PostgreSQL regression test names are strictly alphanumeric plus underscores (e.g., `boolean`, `int4`, `date`, `create_index`), so this is safe for the current use case. A defensive comment noting this assumption would be welcome but is not required.

- **[INFO]** `declare -a ORDERED_TESTS=()` is retained at line 476 — `declare -a` (indexed arrays) is supported since bash 2.x, so this is fine for bash 3.2 portability. Only `declare -A` (associative arrays) needed to be replaced, which was done correctly.

- **[INFO]** Other bash 3.2 considerations — The script also uses `<<<` (here-strings), `[[ =~ ]]` (regex matching), and `read -ra`, all of which are available in bash 3.2. The `BASH_REMATCH` variable is also supported. No remaining bash 4+ features were found.

- **[INFO]** Performance — The PR description correctly notes the `_is_seen` lookup is O(n) per call (string scanning). With typical PostgreSQL regression suites having ~200 tests, the accumulated `_seen_tests` string stays well under a few KiB. This is negligible overhead compared to the actual test execution time.

- **[INFO]** The warning message is correctly sent to stderr and is non-fatal, which aligns with the issue's acceptance criteria ("works on bash 3.2 or fails with a clear error message"). Good UX: warns without blocking.

- **[INFO]** Shell style compliance — 2-space indentation, all variables quoted with `${...}`, `[[ ]]` used throughout, functions use `lower_case` naming, errors go to stderr. Consistent with CLAUDE.md shell style rules.

### Summary

Clean, well-scoped fix. The `declare -A` associative array (the only bash 4+ feature in the script) is replaced with a portable pipe-delimited string and `case`-based lookup. The approach is correct and safe for the domain of PostgreSQL regression test names. CI checks (Lint, Test, Compatibility, Integration Tests, PostgreSQL Regress Tests) all pass. No security, bug, or guideline issues found.

PR #803, opened 2026-04-12
fix(compat): emit rpg:file:line: error location prefix

### LINKED ISSUE #792: rpg does not emit file:line error location prefix
## Problem
When processing SQL files with `-f`, psql prefixes error messages with the file and line number:
```
psql:script.sql:42: ERROR:  syntax error
```

rpg does not emit this prefix — errors just show `ERROR:` without location context.

The test normalizer strips these prefixes (`s/^psql:[^:]*:[0-9]*: //`), masking this difference.

## Acceptance criteria
- rpg emits `rpg:filename:line:` prefix for errors during `-f` file processing
- Remove the error prefix normalization rule from `test-psql-regress.sh`

NikolayS:
## Summary

- Add `current_line_number: Option<u64>` field to `ReplSettings` to track the 1-based line number during file execution
- Emit `rpg:filename:line: ` prefix before error messages during `-f` / `\i` / `\ir` file processing, matching psql's `psql:filename:line: ` behaviour
- Add `eprint_db_error_located()` helper in `output.rs` that prepends the prefix before delegating to `eprint_db_error()`
- Thread the prefix through all error output paths in `execute.rs` (query execution, extended query, gexec, gset, crosstabview) and `mod.rs` (COPY TO/FROM error paths)
- No prefix emitted for `-c` commands, piped stdin, or interactive mode

## Test plan

- [ ] Verify `rpg -f script.sql` emits `rpg:script.sql:N: ERROR:  ...` for SQL errors (where N is the line of the terminating `;`)
- [ ] Verify `rpg -c "SELECT bad"` does NOT emit a file:line prefix
- [ ] Verify interactive mode does NOT emit a file:line prefix
- [ ] Verify `\i` / `\ir` in scripts emit the correct filename and line number
- [ ] Verify nested `\i` correctly restores the outer file's line counter
- [ ] CI regression tests pass (the normalizer strips both `psql:` and `rpg:` prefixes for comparison)

Closes #792

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -59,10 +59,13 @@ pub fn include_file<'a>(
             }
         };
 
-        // Save and update current_file so that nested \ir commands resolve
-        // paths relative to the directory of this file.
+        // Save and update current_file / line counter so that nested \ir
+        // commands resolve paths relative to the directory of this file,
+        // and error messages include the correct `rpg:filename:line:` prefix.
         let prev_file = settings.current_file.clone();
+        let prev_line = settings.current_line_number;
         settings.current_file = Some(path.to_owned());
+        settings.current_line_number = Some(0);
 
         let start_depth = settings.cond.depth();
 
@@ -99,9 +102,10 @@ pub fn include_file<'a>(
             }
         }
 
-        // Restore the previous current_file so callers see the right value
-        // after this include returns.
+        // Restore the previous current_file and line counter so callers
+        // see the right values after this include returns.
         settings.current_file = prev_file;
+        settings.current_line_number = prev_line;
 
         let end_depth = settings.cond.depth();
         if end_depth > start_depth {
--- a/src/output.rs
+++ b/src/output.rs
@@ -1763,6 +1763,25 @@ pub fn eprint_db_error(
     eprint!("{msg}");
 }
 
+/// Print a `tokio_postgres::Error` to stderr with a file location prefix.
+///
+/// Like [`eprint_db_error`] but prepends `rpg:filename:line: ` to the first
+/// line of the error message when `prefix` is `Some`, matching psql's
+/// behaviour during `-f` / `\i` file execution.
+pub fn eprint_db_error_located(
+    prefix: Option<&str>,
+    err: &tokio_postgres::Error,
+    sql: Option<&str>,
+    verbose: bool,
+    terse: bool,
+    sqlstate: bool,
+) {
+    if let Some(pfx) = prefix {
+        eprint!("{pfx}");
+    }
+    eprint_db_error(err, sql, verbose, terse, sqlstate);
+}
+
 /// Format a `PostgreSQL` notice (from `tokio_postgres::error::DbError`) in psql
 /// style, with a colored severity prefix.
 ///
--- a/src/repl/execute.rs
+++ b/src/repl/execute.rs
@@ -609,7 +609,8 @@ pub async fn execute_query(
             if settings.echo_errors {
                 eprintln!("{sql_to_send}");
             }
-            crate::output::eprint_db_error(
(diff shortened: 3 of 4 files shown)


NikolayS:
## REV Review — PR #803

**Verdict:** PASS

### Findings

- [NON-BLOCKING] **Line counter drift during COPY FROM STDIN data consumption.** In `exec_lines`, the `for dl in &mut *lines` loops that consume COPY data lines (both the `\copy` meta-command path around line 2205 and the SQL `COPY ... FROM STDIN` path around line 3069) advance the iterator without incrementing `current_line_number`. After a COPY block with N data lines + the `\.` terminator, subsequent error line numbers will be off by N+1. psql has the same structural pattern (it counts lines only in its main loop), so this is consistent with psql behavior and not a correctness bug per se, but worth noting for future reference.

- [NON-BLOCKING] **Unconverted `eprint_db_error` calls in `describe.rs` and `execute.rs::describe_buffer`.** The `describe.rs` functions (lines 237, 2664, 3000) and `execute.rs::describe_buffer` (lines 2727, 2775) still call plain `eprint_db_error` without the location prefix. These are reachable during file execution via `\d`, `\dt`, `\gdesc`, etc. psql does emit the `psql:file:line:` prefix for these errors too. The impact is minor since describe errors are uncommon, but it means some error paths during `-f` execution will lack the prefix. Similarly, `deallocate_named` (mod.rs line 2128) and `ai_commands.rs` error paths lack the prefix — though the AI commands are rpg-specific so there is no psql equivalence to match.

- [NON-BLOCKING] **Test normalizer still strips both prefixes.** The issue's acceptance criteria says "Remove the error prefix normalization rule from `test-psql-regress.sh`", but the PR keeps lines 191-192 (`s/^psql:...//` and `s/^rpg:...//`). This is actually the correct decision — the normalizer must strip both because psql emits `psql:file:line:` while rpg emits `rpg:file:line:`, and the regression test compares their outputs. However, this means the regression tests cannot verify that rpg actually emits the prefix. A dedicated unit/integration test that checks the prefix is emitted would be valuable.

- [INFO] **Two separate `eprint!` calls for prefix + error.** `eprint_db_error_located` uses `eprint!("{pfx}")` followed by `eprint_db_error(...)` which itself calls `eprint!("{msg}")`. While `eprint!` locks stderr for each call (so there is no interleaving within a single `eprint!`), the prefix and error message are not atomically written as one unit. In practice this is fine since rpg is single-threaded for query execution, but a future refactor could combine them into a single write for robustness.

- [INFO] **`session.rs` calls not converted.** The two `eprint_db_error` calls in `session.rs` (lines 158, 219) are not converted. These handle session-persistence errors that occur outside file-execution context, so the prefix would be `None` anyway. Correctly left unconverted.

- [INFO] **Clean save/restore for nested `\i`/`\ir`.** The `include_file` function in `io.rs` correctly saves both `prev_file` and `prev_line` before setting the new context, and restores both after the include completes. The `exec_file` function in `mod.rs` correctly sets both fields on entry and clears both on exit.

- [INFO] **Prefix correctly suppressed for `-c`, stdin, and interactive mode.** `exec_stdin` does not set `current_line_number`, the interactive REPL never sets it, and `-c` goes through a different path. The `error_location_prefix()` method returns `None` unless both `current_file` and `current_line_number` are `Some`, which is the right guard.

### Summary

Well-structured change that threads file location context through all the main SQL error paths. The prefix format `rpg:filename:line: ` correctly matches psql's pattern. Line counting is 1-based and consistent with psql's behavior (reporting the line of the terminating `;`). Nested `\i`/`\ir` save/restore is correct. The main gap is a handful of secondary error paths (`\d` describe errors, `\gdesc`, `\close_prepared`) that still lack the prefix — these are uncommon but could be addressed in a follow-up.

PR #808, opened 2026-04-12
fix(compat): track standard_conforming_strings for backslash parsing

### LINKED ISSUE #793: standard_conforming_strings=off backslash parsing in tokenizer
## Problem
When `standard_conforming_strings=off`, psql treats `\` inside string literals as an escape character. rpg's tokenizer (`src/highlight.rs`) does not track this GUC, so `\b` inside a string literal is incorrectly treated as a metacommand.

Example: `SELECT 'a\bcd'` with scs=off → rpg errors with `invalid command \b`.

This causes the `strings` regression test to fail in CI Docker environments.

## Acceptance criteria
- rpg tracks `standard_conforming_strings` setting (query on connect, listen for SET)
- Tokenizer respects scs=off for backslash escapes inside string literals
- `strings` regression test passes
- Remove `strings` from `REGRESS_SKIP` in CI

NikolayS:
## Summary

Fixes #793: When `standard_conforming_strings=off`, rpg's tokenizer incorrectly treated backslash inside single-quoted string literals as metacommand prefixes (e.g., `\b` in `SELECT 'a\bcd'`).

- Track `standard_conforming_strings` GUC on connect, reconnect, and mid-session `SET`
- Update tokenizer (`tokenize_scs`) to handle `\x` escapes when SCS=off
- Update inline backslash finder (`find_inline_backslash_ctx`) with SCS-aware string parsing
- Update `get_open_dollar_tag()` and `split_on_backslash_semicolon()` similarly
- Add `E'...'` escape-string support: backslash always escapes regardless of SCS
- Thread SCS state through `RpgHelper` for live syntax highlighting
- 14 new unit tests for SCS-off behavior

**Note:** The `strings` regression test should be removed from `REGRESS_SKIP` in `.github/workflows/checks.yml` once this PR is merged. The workflow file change requires elevated GitHub token permissions (`workflow` scope) that are not available to this agent.

## Files changed

| File | Change |
|------|--------|
| `src/capabilities.rs` | Add `standard_conforming_strings` field + detection |
| `src/highlight.rs` | SCS-aware tokenizer (`tokenize_scs`) + E-string support |
| `src/repl/mod.rs` | SCS-aware inline backslash/dollar-tag/semicolon parsing |
| `src/repl/execute.rs` | Detect mid-session `SET standard_conforming_strings` |
| `src/complete.rs` | Thread SCS to `RpgHelper` highlighter |
| `src/main.rs` | Query SCS at startup (both interactive and non-interactive) |

## Test plan

- [x] All 2006 existing + new unit tests pass (`cargo test`)
- [x] Clean build with no warnings (`cargo build`)
- [ ] `strings` regression test passes in CI (remove from `REGRESS_SKIP`)
- [ ] Verify with live PG: `SET standard_conforming_strings = off; SELECT 'a\bcd';`


NikolayS:
## REV Review — PR #808

**Verdict:** PASS (with non-blocking findings)

---

### Findings

#### 1. `update_scs_if_changed` misses `SET LOCAL` and leading whitespace — NON-BLOCKING

**File:** `src/repl/execute.rs`, lines 1031–1039

The detection heuristic uses `upper.starts_with("SET ")`, which correctly matches `SET LOCAL standard_conforming_strings = off` (since it still starts with `SET `). However, it will miss SQL with leading whitespace or newlines, e.g. `  SET standard_conforming_strings = off`, because `sql_to_send` is not trimmed before the `starts_with` check.

This is a minor robustness gap. The practical impact is low because:
- rpg's REPL typically trims input before sending, and
- The GUC will be re-detected on next reconnect.

**Suggestion:** Consider `upper.trim_start().starts_with(...)` instead of `upper.starts_with(...)`.

#### 2. `find_raw_continuation` not updated with SCS awareness — NON-BLOCKING

**File:** `src/repl/mod.rs`, lines 6946–7019

`find_raw_continuation` has a simple `in_single` toggle on every `'` byte but does not account for backslash escapes (SCS=off) or E-strings. With SCS=off, a `\'` inside a string literal would cause the function to incorrectly toggle `in_single`, potentially miscounting backslash commands.

Impact is low: this function operates on the metacommand continuation portion (after the inline backslash has been found), so string literals with backslash escapes are unlikely to appear in practice.

#### 3. Three `highlight_sql()` callers not updated to pass SCS — NON-BLOCKING

The following call sites still use the non-SCS `highlight_sql()` wrapper (which defaults to `scs=true`):

- `src/markdown.rs:191` — AI/markdown code block highlighting
- `src/repl/ai_commands.rs:1177` — `/ask` SQL echo
- `src/repl/mod.rs:5854` — `\s` history listing

These will highlight string literals using SCS=on semantics even when the server has SCS=off. The visual impact is negligible (highlighting doesn't affect execution), but for full consistency they should call `highlight_sql_scs(...)` with the actual SCS value. These callers may not have easy access to `settings` from their current scope, so this is reasonable to defer.

#### 4. `update_scs_if_changed` not called from `execute_query_extended` — POTENTIAL

**File:** `src/repl/execute.rs`

`update_scs_if_changed` is called from `execute_query` (simple query protocol) but not from `execute_query_extended` (extended query protocol). If a user runs `SET standard_conforming_strings` via `\bind` + extended protocol, the SCS state won't update.

Impact is very low: `SET` via extended protocol with `\bind` is extremely unusual.

#### 5. E-string detection does not check preceding context — INFO

**File:** `src/highlight.rs`, `src/repl/mod.rs`

The E-string check `(bytes[pos] == b'E' || bytes[pos] == b'e') && bytes[pos+1] == b'\''` does not verify that `E` is not part of a larger identifier (e.g., `myvar_e'...'`). However, this is safe because:

- In the tokenizer (`highlight.rs`), the identifier branch consumes the full word first, so by the time `E'` is checked, `pos` is at a token boundary.
- In the `find_inline_backslash_ctx` / `get_open_dollar_tag` / `split_on_backslash_semicolon` byte scanners, an `E` that is part of an identifier would first be consumed character-by-character. However, these scanners do NOT consume identifiers as tokens — they iterate byte-by-byte and check patterns. This means `myvar_e'...'` would see `e` at position N, check `bytes[N+1] == b'\''`, match, and incorrectly enter E-string mode. In practice, an identifier immediately followed by a quote without whitespace is extremely rare in SQL (PostgreSQL's own lexer requires the same `E'` pattern at word boundary). Still, this is a theoretical inconsistency with the tokenizer.

No action needed — documenting for completeness.

#### 6. Default `scs=true` is the safe choice — INFO

The `Default` impl for `DbCapabilities` and the `detect_standard_conforming_strings` fallback both default to `true`, matching PostgreSQL's default since 9.1. This means on error or when connecting to very old PG, backslash is treated as a regular character inside strings (the standard-conforming behavior). This is the correct safe default — it avoids accidentally consuming characters as escapes.

#### 7. Tests are solid but missing one edge case — INFO

The 14 new tests comprehensively cover:
- SCS=on vs SCS=off backslash handling
- Escaped quotes (`\'`) with SCS=off
- Double backslash (`\\`) with SCS=off
- E-strings (both cases, both capitalizations)
- Regression test (no false `BackslashCmd` tokens)
- Default equivalence (`tokenize()` == `tokenize_scs(true)`)
- SCS-aware `find_inline_backslash_scs` tests

One optional addition: a test for `E'hello\\'world'` (E-string with escaped backslash before a quote) to verify the backslash-then-quote sequence inside E-strings. Current tests only cover `E'a\bcd'`.

#### 8. Reconnect paths are properly covered — INFO

SCS is re-detected at all reconnect points:
- `run_readline_loop` reconnect (line 6613)
- `run_dumb_loop` reconnect — three separate reconnect sites (lines 6733, 6768, 6795)
- `main.rs` startup (both interactive and non-interactive paths)
- `RpgHelper` is updated with fresh SCS after reconnect

This is thorough and correct.

---

### Summary

This is a well-executed change that correctly tracks `standard_conforming_strings` across connect, reconnect, and mid-session `SET`, threading the state through all three string-parsing code paths (tokenizer, inline-backslash finder, backslash-semicolon splitter) and the syntax highlighter. The E-string handling is correct. The safe default of `scs=true` on error is the right call.

The non-blocking findings (leading-whitespace in SET detection, `find_raw_continuation` not SCS-aware, a few `highlight_sql` callers) are minor robustness gaps that can be addressed in follow-up work. No security or correctness blockers found.

NikolayS:
Fixed the CI `cargo fmt` failure. The formatting issues were in 5 files introduced by the SCS backslash parsing commit:

- `src/capabilities.rs` — function signature line wrapping
- `src/complete.rs` — multi-line function call formatting
- `src/highlight.rs` — conditional expression line wrapping
- `src/repl/execute.rs` — function signature line wrapping
- `src/repl/mod.rs` — conditional expressions and `if let` formatting

Applied `cargo fmt`, verified clean build and no new clippy warnings. Pushed as a new commit: `fix: apply cargo fmt formatting (#793)`.

NikolayS:
Fixed all 3 clippy lints:

1. **`struct_excessive_bools`** on `RpgHelper` — added `#[allow(clippy::struct_excessive_bools)]` since the 4 bools (`highlight`, `standard_conforming_strings`, `completion_enabled`, `dropdown_completion_enabled`) are genuinely independent toggle flags.

2. **`too_many_lines`** on `run_dumb_loop` (108/100) — extracted `apply_dumb_reconnect()` helper to deduplicate the reconnect-handling block that appeared twice (once for leading `\cmd`, once for inline `\cmd`). Now well under the limit.

3. **`too_many_lines`** on `find_inline_backslash_ctx` (111/100) — extracted `try_parse_dollar_tag()` helper for dollar-quote detection, converted the trailing if-chain to a `match`, and condensed variable declarations. Also reused the new helper in `split_on_backslash_semicolon` for consistency.

All checks pass locally: `cargo clippy` (clean), `cargo fmt --check` (clean), `cargo build` (success).

PR #813, opened 2026-04-13
docs: add red/green TDD section to CLAUDE.md

NikolayS:
## Summary
- Adds a "Testing approach — red/green TDD" section to CLAUDE.md
- Bug fixes must include a test that would have caught the bug
- New features should write the test first when practical
- For psql compat work, the regression test diff serves as the "red" phase


NikolayS:
Postponed — will revisit later.

Issue #824, opened 2026-04-21
Comprehensive review: findings across docs, code quality, security, and CI (v0.11.0)


NikolayS:
## Scope & method

End-to-end review of rpg at `765ebdb` (v0.11.0). Parallel investigations covered:

- **Build & lint**: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo build --release`, `cargo test --lib --bins`
- **Docs/spec drift**: README, Cargo.toml, CONTRIBUTING, CHANGELOG, `docs/`, `examples/`, `demos/`
- **Code quality**: module sizing, error handling, duplication, API design, test coverage, dead-code allows
- **Security**: secret handling, SQL/command injection, TLS defaults, SSH host keys, Lua sandboxing, project-config trust, file permissions, AI data exposure

### Build/test results (all green)

| Check | Result |
|---|---|
| `cargo fmt --check` | PASS |
| `cargo clippy --all-targets --all-features -- -D warnings` | PASS |
| `cargo test --lib --bins` | **2063 passed, 0 failed, 4 ignored** |
| `cargo build --release` | PASS (12 MB binary) |
| `target/release/rpg --version` | Works |

So the codebase is in good shape mechanically. The findings below are correctness, hygiene, and posture issues — not broken builds.

---

### H1. `Cargo.toml` repository URL points at the old project name

`Cargo.toml:8` → `repository = "https://github.com/NikolayS/project-alpha"`. Actual repo is `NikolayS/rpg`. This URL is shipped into crate metadata and shown on crates.io / `cargo search`.

**Fix:** `repository = "https://github.com/NikolayS/rpg"`

### H2. Rust minimum version mismatch between README and `Cargo.toml`

- `README.md:35` — "requires Rust 1.85+"
- `Cargo.toml:5` — `rust-version = "1.82"`
- CI lint job uses `toolchain: stable` without pinning a minimum (`.github/workflows/checks.yml:24`).

One of them is wrong. Pick the actual MSRV, document it, and pin CI to it (so MSRV breakage is caught).

### H3. `CONTRIBUTING.md` still uses the old project name

`CONTRIBUTING.md:17–18` clones `git@github.com:NikolayS/project-alpha.git` and `cd project-alpha`. Same rename miss as H1 — any new contributor hits a 404.

### H4. CI tests mostly against PG 16; only the regress job uses PG 18

[...]

PR #817, opened 2026-04-13
fix(security): break cleartext-logging taint chain

NikolayS:
## Summary

- Fix CodeQL "cleartext logging of sensitive information" (High severity) alert on `src/main.rs:928`
- Introduce `ConnDisplayInfo` — a password-free, borrow-only subset of `ConnParams` — so that `connection_info()` and `reconnect_info()` never receive the password field
- This cleanly breaks the taint chain from `resolve_password()` → `ConnParams.password` → `println!` that CodeQL was flagging
- The previous code passed the full `ConnParams` struct (which contains `password: Option<String>`) to display functions; although the password was never accessed, CodeQL's taint analysis flagged the struct-level data flow

## Test plan

- [x] `cargo test` — all 2025 tests pass
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [ ] CodeQL "Analyze (rust)" check passes on this PR

https://claude.ai/code/session_01H65WbzpPvcBgsWYkx5V9YK

PR #831, opened 2026-04-25
fix(security): trust check for project-level .rpg.toml (#824 H5)

NikolayS:
## Summary

Fixes finding **H5** from #824. A `.rpg.toml` discovered by walking CWD up to `$HOME` is loaded with no permission or ownership check. Because `PROMPT1` and friends flow through backtick-shell expansion in the REPL (`src/repl/mod.rs:282`), **cloning an untrusted repo and starting rpg inside it can execute arbitrary shell on first prompt render**.

## Threat model

1. Attacker controls a Git repo (or any directory rpg might be invoked in).
2. Repo contains a `.rpg.toml` setting `PROMPT1 = "\`evil-cmd\`db=> "`.
3. Victim clones the repo and runs `rpg`. rpg walks up from CWD, finds `.rpg.toml`, loads it. The REPL renders the prompt and executes `evil-cmd` via `sh -c`.

Analogous tools (direnv `.envrc`, VS Code workspace settings) require explicit trust for exactly this reason. pgpass already has a permission check at `src/connection.rs:1566–1582`; the far more powerful `.rpg.toml` did not.

## Minimum-viable fix

New `check_project_config_trust()` in `src/config.rs`, modelled on the pgpass check:

- File must be **owned by the current uid**.
- Mode must satisfy `& 0o022 == 0` (no group/other **write**).

Untrusted files are **skipped with a stderr warning** and an actionable `chmod 0600` hint; the REPL still starts. User-level config under `~/.config/rpg/` is implicitly trusted and **unaffected** (matches how psql treats `~/.psqlrc`).

The check is `#[cfg(unix)]`; on Windows the loader proceeds because POSIX modes are not portably exposed — same gating as the pgpass check.

## Deferred work

The richer TOFU-style allowlist at `~/.config/rpg/trusted_project_configs` (direnv-style opt-in trust) suggested in #824 is **out of scope** for this PR; it is left as a `TODO(#824 H5)` in `check_project_config_trust`'s docstring. Implementing this minimum-viable check first lets us ship the fix immediately and design the allowlist UX deliberately.

H6 (env scrubbing on the backtick subshell) is being addressed in a parallel PR.

## Test plan

Red/green TDD per CLAUDE.md.

- [x] **Red**: `config::tests::load_project_config_skips_world_writable_file` — creates a temp dir with a world-writable `.rpg.toml`, points CWD at it, asserts `load_project_config()` does not load the file. Verified failing on `main` (the world-writable config was loaded).
- [x] **Green**: same test passes after the fix.
- [x] `cargo test --lib --bins` — 2064 passed, 0 failed (1 new test added).
- [x] `cargo fmt --check` — clean.
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo check --target wasm32-unknown-unknown` — builds.

Refs #824.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -808,18 +808,17 @@ fn check_project_config_trust(path: &Path) -> Result<(), String> {
     {
         use std::os::unix::fs::MetadataExt;
         let meta = std::fs::metadata(path).map_err(|e| format!("cannot stat: {e}"))?;
-        let uid = unsafe { libc::geteuid() };
-        if meta.uid() != uid {
-            return Err(format!(
-                "owned by uid {} (current uid {uid}); refusing to load",
-                meta.uid()
-            ));
+        // Compare uids without including either value in the error message:
+        // CodeQL's cleartext-logging rule treats `geteuid()` results as
+        // sensitive and flags them when they reach a logger (mirrors #817's
+        // taint-chain break for ConnParams).  The diagnostic value of
+        // showing the numeric uids is low — the user can `ls -l` the file.
+        if meta.uid() != unsafe { libc::geteuid() } {
+            return Err("not owned by the current user; refusing to load".into());
         }
-        let mode = meta.mode() & 0o777;
-        if mode & 0o022 != 0 {
+        if meta.mode() & 0o022 != 0 {
             return Err(format!(
-                "mode {mode:04o} is group/other writable; \
-                 run `chmod 0600 {}` to fix",
+                "group/other writable; run `chmod 0600 {}` to fix",
                 path.display()
             ));
         }
--- a/src/config.rs
+++ b/src/config.rs
@@ -781,6 +781,55 @@ pub fn find_project_config(_start_dir: &Path) -> Option<PathBuf> {
     None
 }
 
+/// Verify a project-level `.rpg.toml` is safe to load (#824 H5).
+///
+/// Project configs can set fields like `PROMPT1` that flow through the REPL's
+/// backtick subshell expansion (`src/repl/mod.rs:282`), so a world-writable or
+/// other-owned `.rpg.toml` is a code-execution vector when running rpg inside
+/// an untrusted clone.  Mirror the pgpass check at
+/// `src/connection.rs:1566–1582`: require current-user ownership and no
+/// group/other write bits.
+///
+/// Returns `Ok(())` when the file is trusted, or `Err(reason)` describing the
+/// violation.  On non-unix platforms the OS does not expose POSIX ownership
+/// or mode bits in a portable way, so the check is a no-op there — matching
+/// how the pgpass check is also gated `#[cfg(unix)]`.
+///
+/// User-level configs under `~/.config/rpg/` are *not* run through this
+/// check: the home directory is implicitly trusted, same as psql treats
+/// `~/.psqlrc`.
+///
+/// TODO(#824 H5): follow up with a TOFU-style allowlist at
+/// `~/.config/rpg/trusted_project_configs` (direnv-style opt-in trust) so
+/// users can knowingly load configs that fail this minimum check.
+#[cfg(not(target_arch = "wasm32"))]
+fn check_project_config_trust(path: &Path) -> Result<(), String> {
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::MetadataExt;
+        let meta = std::fs::metadata(path).map_err(|e| format!("cannot stat: {e}"))?;
+        let uid = unsafe { libc::geteuid() };
+        if meta.uid() != uid {
+            return Err(format!(
+                "owned by uid {} (current uid {uid}); refusing to load",
+                meta.uid()
+            ));
+        }
+        let mode = meta.mode() & 0o777;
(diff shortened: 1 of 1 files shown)


NikolayS:
Addressed CodeQL cleartext-logging finding (security/code-scanning/140) in 6e5b1a9.

The flagged taint flow was `meta.uid()` / `geteuid()` → `format!` → `rpg_eprintln!`. Same shape as #817 fixed in `connection.rs` last week — uid values are sensitive in CodeQL's taint model.

Fix: drop the numeric uids from the error message. Diagnostic value was low (`ls -l` shows ownership), and the trust check itself is unchanged. Mode error keeps the `chmod 0600` hint, which is the actually useful bit.

Existing test `load_project_config_skips_world_writable_file` still passes — it asserts on rejection, not the error string.

NikolayS:
## REV review — PR #831

Tight, well-scoped fix for #824 H5: trust check on project `.rpg.toml` mirroring the pgpass shape, gated `#[cfg(unix)]`, with a real red→green→fix commit sequence and a passing world-writable test.

### Findings

- **INFO** security (`src/config.rs:819–824`): `POSTGRES.md` adjacent to an untrusted `.rpg.toml` is still read unconditionally and flows into AI context. Not a shell-exec vector and out of scope for H5, but the same hostile-clone threat model applies — consider a follow-up issue.
- **INFO** security (`src/config.rs:806`): `std::fs::metadata` follows symlinks, so a symlink to attacker-owned content would still face the target file's mode check. Matches `pgpass` posture (`src/connection.rs:1566–1582` is also `metadata`-based) — consistent. Worth a one-line note in a follow-up audit.
- **INFO** security (non-unix): the loader proceeds without any check on Windows/wasm. The TODO mentions the TOFU follow-up; H6's prompt-expansion neutralisation is the actual defence-in-depth on those platforms. Documented adequately for now.
- **NON-BLOCKING** bugs (`src/config.rs:828`): `let _ = path;` after the `#[cfg(unix)]` block is a clippy silencer for the non-unix arm. Cleaner pattern: rename the parameter to `_path` on non-unix via `#[cfg_attr(not(unix), allow(unused_variables))]`. Cosmetic.
- **INFO** security spec: `& 0o022` (write-only) is the correct mask for the threat — read access does not leak secrets from a project config, and stricter `& 0o077` would reject many real repos with group-readable trees.
- **INFO** bugs: only one call site (`src/main.rs:748`) reaches `load_project_config`; the only `read_to_string` of `.rpg.toml` is inside the post-trust branch. No bypass path.
- **INFO** tests: `load_project_config_skips_world_writable_file` passes; uses an `h5_canary` named query and asserts both `config_path.is_none()` and the entry is absent — would not pass without the fix. No happy-path test for a 0o600 file in this PR (a follow-up improvement).
- **INFO** guidelines: three commits (`ef39044`, `0e4615f`, `6e5b1a9`), Conventional Commits, imperative, < 72 chars; rustdoc on `check_project_config_trust` cites #824, the pgpass parallel, the REPL backtick line, and the deferred TOFU TODO.
- **INFO** CodeQL: the third commit (`6e5b1a9`) addresses an automated cleartext-logging finding by dropping the numeric uids from the error string — same shape as the previous fix at #817.

### Verdict

**PASS**

---

## End-to-end test evidence

### Unit-test evidence

```
$ cargo test --bins load_project_config_skips_world_writable_file
running 1 test
test config::tests::load_project_config_skips_world_writable_file ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2067 filtered out; finished in 0.00s
```

### Live end-to-end evidence — negative case (world-writable, rejected)

Built a fresh `rpg` from this branch and ran it from a directory containing a 0o666 `.rpg.toml`:

```
$ ls -la /tmp/rpg-h5-evidence/.rpg.toml
-rw-rw-rw-@ 1 nik wheel 125 Apr 25 20:33 /tmp/rpg-h5-evidence/.rpg.toml

$ cd /tmp/rpg-h5-evidence && rpg -h does-not-exist.invalid -d test_db
rpg: warning: ignoring project config "/private/tmp/rpg-h5-evidence/.rpg.toml": group/other writable; run `chmod 0600 /private/tmp/rpg-h5-evidence/.rpg.toml` to fix
rpg: connection to server at "does-not-exist.invalid", port 5432 failed: failed to lookup address information: nodename nor servname provided, or not known
```

The trust check fires with the exact `chmod 0600 <path>` remediation hint. Loader falls through to `ProjectConfig::default()` (no abort) and continues to the connection step.

### Live end-to-end evidence — positive case (0o600, loaded)

After fixing perms:

```
$ chmod 0600 /tmp/rpg-h5-evidence/.rpg.toml
$ ls -la /tmp/rpg-h5-evidence/.rpg.toml
-rw-------@ 1 nik wheel 125 Apr 25 20:33 /tmp/rpg-h5-evidence/.rpg.toml

$ cd /tmp/rpg-h5-evidence && rpg -h does-not-exist.invalid -d test_db
rpg: connection to server at "does-not-exist.invalid", port 5432 failed: ...
```

No warning — the config was accepted and the loader proceeded without complaint.

